PR #5, opened 2022-09-23
Feat/mvp docker support

nravic:
Thin shell around docker sdk checkpointing. Turns out you can leave a container running!


nravic (commented):
lgtm

PR #7, opened 2022-09-25
fix: client and daemon cleanup

nravic:


nravic (commented):
Added some fluff to set up direct commands from an orchestrator.

PR #29, opened 2023-02-22
Feat/predump signal

nravic:
- Added a predump signal for users to bypass CRIU with 
- Some fluff/plans for intercepting CUDA calls


nravic (commented):
lgtm boss :+1:

PR #43, opened 2023-06-26
fix: rewrite of restore and dump commands

nravic:
- MVP for filesystem checkpointing, open files are sent as part of a larger checkpoint 
- Pytorch checkpointing more robust 
- Restore command rewritten for ease of use w/ commandline


nravic (commented):
lgtm bossman

PR #46, opened 2023-07-10
fix: new unzip function for dealing with subdirectories

nravic:


nravic:
subdirectories weren't being properly inflated, fixed

Issue #64, opened 2023-08-09
[CED-78] Checkpoint storage abstractions


nravic:
With the minio server in place, we need to be able to distinguish between a NATS store for storage and the minio server - based on config. 

Abstract current implementation out into an interface and test against `play.minio.com`.  

<sub>From [SyncLinear.com](https://synclinear.com) | [CED-78](https://linear.app/cedana/issue/CED-78/checkpoint-storage-abstractions)</sub>

PR #65, opened 2023-08-09
CED-78: Checkpoint storage abstractions

nravic:
We want interop between storing checkpoints in minio & in NATS (for self-serve). With all the testing work being done as well, it's useful to be able to mock it out too. 

https://github.com/cedana/cedana/issues/64


bsmithai (approved):
This looks good, simple abstraction

PR #71, opened 2023-08-18
CED-77: Running tasks from daemon (vs over SSH) 

nravic:


bsmithai (commented):
looks good so far!

bsmithai (approved):

cmd/client-daemon.go
```diff
@@ -87,9 +78,29 @@ var clientDaemonCmd = &cobra.Command{
 
 		c.logger.Info().Msgf("daemon started at %s", time.Now().Local())
 
+		// create a subscription to NATS commands from the orchestrator first
 		go c.subscribeToCommands(300)
+
+		if pid == 0 {
+			// we retry startJob up to 5 times until it succeeds everytime we get a new serverCommand
+			var task string = c.config.Client.Task
+			for i := 0; i < 5; i++ {
```
wfoy, line 87: what happens if this doesn't succeed within 5 iterations? Is that possible?

cmd/client-daemon.go
```diff
@@ -103,33 +103,87 @@ var clientDaemonCmd = &cobra.Command{
...
+	// so there's no real concerns here about executing the task without proper validation
+
+	// we want the task to run detached so we can checkpoint it
+	cmd := exec.Command("/bin/sh", "-c", task)
+	cmd.SysProcAttr = &syscall.SysProcAttr{
+		Setsid: true,
+	}
+
+	cmd.Stderr = os.Stderr
+	cmd.Stdout = os.Stdout
```
bsmithai, line 149: Do we not need to throw stdin & stdout into /dev/null?

nravic: yes we do, good catch

nravic: fixed!

bsmithai (approved):

bsmithai:
Amazing!

PR #74, opened 2023-08-30
CED-81: Streaming task logs 

nravic:
Now that we exec tasks straight from the daemon, we can super easily stream logs. 

This absolutely needs a rate limiter, and some figuring out of retention for NATS (so we don't crowd a cluster with persistent logs).


cmd/client-dump.go
```diff
@@ -251,6 +255,26 @@ func (c *Client) copyOpenFiles(dir string) error {
 	return nil
 }
 
+func (c *Client) checkFileLocks(pid int32) bool {
+	for _, fd := range c.Process.OpenFds {
+		info, err := c.fs.ReadFile(fmt.Sprintf("/proc/%s/fdinfo/%s", strconv.Itoa(int(pid)), strconv.FormatUint(fd.Fd, 10)))
```
bsmithai, line 260: Can we use /proc/locks for a better algo?

nravic: just ended up removing the function for now, wasn't really using it anyway

bsmithai (approved):
Looks good!

PR #77, opened 2023-09-14
CED-112: Cleanup & Docker image for benchmarking

nravic:
- Added a feature flag for log forwarding
- Some basic scaffolding for tcp checkpoint testing 
- Added a dockerfile for benchmarking 
- Cleanup of some benchmarking code, can only now exec benchmarks from inside the test/benchmarks folder 
- Changed how RunTask execs a program, use cmd.Exec for better control


bsmithai (approved):
looks good!

PR #79, opened 2023-09-18
CED-139: Reorganize into RPC 

nravic:
Bit of a refactor, moving code around so core capabilities can only be called via RPC. Daemon is now also more of a daemon, and listens on /tmp/cedana.sock. 

Still a lot of features to implement back, but purpose of this ticket is to gain back parity with what it was like pre-refactor.


bsmithai (approved):
lgtm

PR #81, opened 2023-09-22
fix: better cli commands for dump and restore (containerd-runc-process)

bsmithai:
Changes:
- Made dump and restore root commands:
- Commands look like: dump process, dump containerd, dump runc
- Same for restore

### DIFF
--- a/cmd/cli.go
+++ b/cmd/cli.go
@@ -89,7 +89,7 @@ var dumpProcessCmd = &cobra.Command{
 
 var dumpCmd = &cobra.Command{
 	Use:   "dump",
-	Short: "Manually checkpoint a running process to a directory",
+	Short: "Manually checkpoint a process or container to a directory: [process, runc (container), containerd (container)]",
 }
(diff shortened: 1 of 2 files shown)


cmd/cli.go
```diff
@@ -184,6 +188,11 @@ var runcRestoreCmd = &cobra.Command{
 var restoreCmd = &cobra.Command{
 	Use:   "restore",
 	Short: "Manually restore a process from a checkpoint located at input path",
```
nravic, line 190: Change the short description in the parent commands to give a hint for the options? Something like "manually checkpoint [process, runc (container), containerd (container)]"

nravic (commented):
Looks good, runcRestore still hardcodes your path btw

cmd/cli.go
bsmithai: Oh shoot good catches thanks

nravic (approved):

PR #86, opened 2023-09-27
Feat/ced 41

bsmithai:
Design doc:

https://www.notion.so/cedana/Cedana-Market-e2e-22a5e7e443ad4467a8f9152584a48076


services/server/server.go
```diff
@@ -0,0 +1,268 @@
...
+	// Create a buffer to read the file data into
+	buffer := new(bytes.Buffer)
+	_, err = buffer.ReadFrom(file)
+	if err != nil {
+		fmt.Println("Error reading file data:", err)
+		return err
+	}
+
+	httpClient := &http.Client{}
+	url := os.Getenv("CHECKPOINT_SERVICE_URL") + "/checkpoint/6291dc64-289f-4744-9aa6-2a382b0a9a30/upload/" + uploadResp.UploadID
```
avalonbits, line 105: you need to add the /part/{part_number}  as well.

avalonbits (changes requested):

api/server.go
```diff
@@ -0,0 +1,408 @@
...
+	err := s.Client.Dump(args.Dir, args.PID)
+	if err != nil {
+		return nil, err
+	}
+
+	switch args.Type {
+	case task.DumpArgs_SELF_SERVE:
+		// TODO BS: This is a hack for now
+	case task.DumpArgs_MARKET:
+		file, err := os.Open(s.Client.CheckpointDir + ".zip")
```
nravic, line 77: CheckpointDir should go and be replaced by queries to the db for the checkpoint, makes this interface a lot cleaner

cmd/cli.go
```diff
@@ -254,81 +356,40 @@ var startTaskCmd = &cobra.Command{
...
+		taskArgs := &task.StartTaskArgs{
 			Task: args[0],
-			ID:   args[1],
-		}
-
-		var resp api.StartTaskResp
-		err = cli.conn.Call("CedanaDaemon.StartTask", a, &resp)
-		if err != nil {
-			return err
+			// ID:   args[1],
```
nravic, line 361: passing ID to startTask for args needs to come back - it's a simple way to attach a process to a jobID for an orchestrator

nravic (approved):
lgtm, will correct issues as they come up

PR #87, opened 2023-09-28
CED-111: TCP restore benchmarking 

nravic:


nravic:
Added a fix for CED-147 as well by running cmd.Wait in a goroutine.

PR #93, opened 2023-10-17
Feat/cli 52

bsmithai:


cmd/cli.go
```diff
@@ -366,6 +278,63 @@ var containerdDumpCmd = &cobra.Command{
 	},
 }
 
+var gpuDumpCmd = &cobra.Command{
```
nravic, line 281: can you move these to the process checkpoint/restore comment block? And then also add a note about how this is for processes only right now.

nravic (approved):
small comment for readability

cmd/cli.go
bsmithai: Done!

PR #94, opened 2023-10-18
Miscellanous fixes 

nravic:
- Open/close bolt dbs with each transaction. Keeping a connection open prevents other parts of the code from accessing the db. 
- Restore directly from an ID instead of a path. 
- Moved away from passing a directory to restore to a direct path. 
- Better differentiation b/ween a local restore (from a file on disk) and a remote restore (from a file in the cloud)


Brandonsmith738 (approved):
lgtm

PR #95, opened 2023-10-18
Feat/cli 56

bsmithai:
What this solves:

- issue with container id
- patched config.dump to include new directories

What needs done: 

/run/containers patch, so far we just do a /var/lib patch


api/dump.go
```diff
@@ -24,6 +24,11 @@ const (
 	sys_pidfd_getfd       = 438
 )
 
+type Bundle struct {
```
nravic, line 27: Can you add a comment about what a Bundle is/does?

api/restore.go
```diff
@@ -215,17 +206,61 @@ func patchPodmanRestore(opts *container.RuncOpts) error {
 
 func (c *Client) RuncRestore(imgPath, containerId string, opts *container.RuncOpts) error {
 
+	bundle := Bundle{Bundle: opts.Bundle}
+
+	isPodman := checkIfPodman(bundle)
+
+	if isPodman {
```
nravic, line 213: Also a comment explaining what's happening here w/ the rsync

cmd/client-daemon.go
```diff
@@ -63,7 +63,7 @@ var startDaemonCmd = &cobra.Command{
 
 		logger.Info().Msgf("daemon started at %s", time.Now().Local())
 
-		go startgRPCServer()
+		startgRPCServer()
```
nravic, line 66: why was this removed as a goroutine? IIRC the daemon won't work otherwise because startgRPCServer is blocking

nravic (commented):
lgtm, some things I don't understand can be resolved with better comments

cmd/client-daemon.go
Brandonsmith738: I just forgot to re-add this. I drop go daemon whenever I want to debug rpc

bsmithai: Forgot to re-add

nravic (approved):
lgtm

PR #98, opened 2023-10-19
Feat/cli 57

bsmithai:


api/server.go
```diff
@@ -109,26 +120,29 @@ func (s *service) Dump(ctx context.Context, args *task.DumpArgs) (*task.DumpResp
...
 
 		return &task.DumpResp{
-			Error: fmt.Sprintf("Dumped process %d to %s, multipart checkpoint id: %s", args.PID, args.Dir, multipartCheckpointResp.UploadID),
+			Message: fmt.Sprintf("Dumped process %d to %s, multipart checkpoint id: %s", args.PID, args.Dir, multipartCheckpointResp.UploadID),
 		}, nil
 	}
 
 	return &task.DumpResp{
-		Error: fmt.Sprintf("Dumped process %d to %s", args.PID, args.Dir),
+		Message: fmt.Sprintf("Dumped process %d to %s", args.PID, args.Dir),
```
nravic, line 145: I think you should try stuffing this data into the errorcode metadata like in the blog post, instead kind of like this: 

```
desc := "The username must only contain alphanumeric characters"
v := &errdetails.BadRequest_FieldViolation{
    Field: "username",
    Description: desc,
}
st, err := st.WithDetails(v)
```

It'll help us I think avoiding polluting the grpc w/ random messages and also makes error checking on the other side easier

cmd/client-daemon.go
```diff
@@ -33,42 +33,42 @@ var startDaemonCmd = &cobra.Command{
...
+		// 	logger.Fatal().Msg("Could not find cedana executable")
+		// }
+
+		// ctx := &gd.Context{
+		// 	PidFileName: "/tmp/cedana.pid",
+		// 	PidFilePerm: 0o644,
+		// 	LogFileName: "/tmp/cedana-daemon.log",
+		// 	LogFilePerm: 0o664,
+		// 	WorkDir:     "./",
+		// 	Umask:       027,
```
nravic, line 47: was everything here accidentally commented out?

api/server.go
bsmithai: I can see where I can use this elsewhere but that return statement is where err is nil and there was a successful dump

nravic: oh yup, picked a bad example

nravic (approved):

PR #101, opened 2023-10-23
Feat/tcp script

nravic:


bsmithai (commented):
lgtm, never seen subprocess management in python before so kinda cool to see.

PR #102, opened 2023-10-23
fix: properly grab task to run from config

bsmithai:
Quick fix to get tasks running based off config json that is created from instantiation of client


nravic (approved):
lgtm

PR #107, opened 2023-11-10
feat/CED-173: Better benchmarking

nravic:
Encapsulates a bunch of changes. 

Added: 
- A working directory to specify to execTask 
- get out of go benchmarking tooling and into python (ugh)
- replace go benchmarking with running a pprof server now instead to profile checkpoint/restore. helps segment different ops 
- some coarse data generated from benchmark.py 

Todo (will be further tickets) 
- Fix containerization 
- push to bq or datastore for more analysis 
- analyze generated pprof files using the pprof proto


api/server.go
```diff
@@ -345,9 +344,22 @@ func (s *service) runTask(task string) (int32, error) {
...
+	if workingDir != "" {
+		cmd.Dir = workingDir
+	}
+
 	cmd.Stdin = nullFile
-	cmd.Stdout = w
-	cmd.Stderr = w
+	if logOutputFile == "" {
+		// default to /var/log/cedana-output.log
+		logOutputFile = "/var/log/cedana-output.log"
```
rst0git, line 354: It might be better to use constants for default values

cmd/client-daemon.go
```diff
@@ -80,7 +85,7 @@ var stopDaemonCmd = &cobra.Command{
 	RunE: func(cmd *cobra.Command, args []string) error {
 		// kill -9 daemon
 		// read from PID file
-		pidFile, err := os.ReadFile("/tmp/cedana.pid")
+		pidFile, err := os.ReadFile("/var/log/cedana.pid")
```
rst0git, line 88: `/run/cedana.pid` might be more appropriate location for the pid file

nravic: good call! fixed

bsmithai (approved):
lgtm

PR #134, opened 2024-02-02
Feat/ced 190

wfoy:


cmd/client-daemon.go
```diff
@@ -66,6 +68,18 @@ var startDaemonCmd = &cobra.Command{
 			go startProfiler()
 		}
 
+		if os.Getenv("CEDANA_GPU_ENABLED") == "true" {
+			err = pullGPUBinary("gpucontroller", "./gpu-controller")
```
nravic, line 72: let's set filepath for both to /usr/local/bin/cedana-gpu-controller & /usr/local/lib/libcedana-gpu.so

nravic (commented):
There's some artifacts from when I was messing with this PR - I'll clean those up.

PR #135, opened 2024-02-02
Logfile redirection

nravic:


api/restore.go
```diff
@@ -51,43 +69,65 @@ func (c *Client) prepareRestore(opts *rpc.CriuOpts, args *task.RestoreArgs, chec
...
 	}
 
-	// check open_fds. Useful for checking if process being restored
-	// is a pts slave and for determining how to handle files that were being written to.
-	// TODO: We should be looking at the images instead
 	open_fds := checkpointState.ProcessInfo.OpenFds
-	var isShellJob bool
+
+	// create logfile for redirection
+	file, err := os.Create("/var/log/cedana-output-restored.log")
```
bsmithai, line 91: What happens if you c/r again on the restored process? It will log to the same restored log right?

nravic: hmm yeah good point, maybe I'll throw a timestamp in

nravic: fixed

PR #136, opened 2024-02-03
Feat/no daemon

nravic:
go-daemon -> no-daemon 

- Removed go-daemon and replaced w/ systemd 
- Added dev script 
- Some more permissions stuff

### DIFF
--- a/test/benchmarks/Dockerfile
+++ b/test/benchmarks/Dockerfile
@@ -17,9 +17,6 @@ RUN git clone https://github.com/cedana/cedana && mkdir ~/.cedana
 WORKDIR /cedana 
 
 ENV USER="root"
-RUN git fetch && git checkout feat/no-daemon
-RUN go build && ./cedana bootstrap 
-
 ## set up bq env and run benchmarks
 ENV CEDANA_PROFILING_ENABLED=true 
 ENV CEDANA_REMOTE=true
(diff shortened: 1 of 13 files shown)


test/benchmarks/Dockerfile
```diff
@@ -29,5 +27,7 @@ RUN mkdir benchmark_results
 RUN pip install protobuf==3.20.1 requests psutil google-cloud-bigquery torch
 RUN cd test/benchmarks && protoc --python_out=. profile.proto
 
-## start daemon and benchmark 
-CMD ["sh", "-c", "./cedana daemon start && python3 test/benchmarks/benchmark.py"]
\ No newline at end of file
+
+RUN go build && ./cedana bootstrap 
+
+CMD ["sh", "-c", "sudo -E ./cedana daemon start & sudo python3 test/benchmarks/benchmark.py"]
```
bsmithai, line 33: This still has cedana daemon start ?

bsmithai (changes requested):

test/benchmarks/Dockerfile
nravic: yep - you can start it manually like this; it in effect just runs the rpc server. Take a look at the build-start-daemon file as well or client-daemon.go to see what I mean

PR #140, opened 2024-02-22
Feat: Switch to Otel

nravic:
- Use OTel for metrics 
- Readme cleanup 
- Benchmarks updated to use OTel 
- Removed timers 
- Better ctx propagation


bsmithai (approved):
lgtm

PR #142, opened 2024-03-01
Feat/gce 99

bsmithai:
## Additions:
- Cedana Helper
- EKS Bug fixes


api/services/cts.go
```diff
@@ -17,12 +16,12 @@ type ServiceClient struct {
...
 }
 
-func NewClient(addr string) *ServiceClient {
+func NewClient(addr string) (*ServiceClient, error) {
 	var opts []grpc.DialOption
 	opts = append(opts, grpc.WithTransportCredentials(insecure.NewCredentials()))
 	taskConn, err := grpc.Dial(addr, opts...)
 	if err != nil {
-		log.Fatalf("fail to dial: %v", err)
+		return nil, err
```
nravic, line 24: why fail silently now?

cmd/cedana-helper/install.sh
```diff
@@ -0,0 +1,34 @@
+#!/bin/bash
+
+# Check whether git is already installed
+if ! command -v git &> /dev/null; then
+    yum install -y git
```
nravic, line 5: why yum? should we check the base OS and decide the package manager that way?

cmd/cedana-helper/install.sh
```diff
@@ -0,0 +1,34 @@
...
+
+# Add export statement to ~/.bashrc
+echo "export PATH=$PATH:/usr/local/go/bin" >> /root/.bashrc
+
+# Install required packages
+yum install -y libnet-devel protobuf protobuf-c protobuf-c-devel protobuf-compiler protobuf-devel protobuf-python libnl3-devel libcap-devel
+yum group install -y "Development Tools"
+
+# Clone CRIU repository and build
+git clone https://github.com/checkpoint-restore/criu.git
```
nravic, line 24: what's the frequency of this install.sh operation? can you time it? concerned building from source is going to take too long

test.yaml
```diff
@@ -0,0 +1,113 @@
...
+      interval: 1s
+      timeout: 5s
+      retries: 100
+    restart: always
+    volumes:
+      - ./redis.conf:/usr/local/etc/redis/redis.conf
+    ports:
+      - "6379:6379"
+  sqs:
+    container_name: mesocast-sqs
```
nravic, line 18: hell yeah using mesocast is a sick idea

test.yaml
```diff
@@ -0,0 +1,113 @@
+version: '2.3'
+
```
nravic, line 2: can you toss a comment in here about the mesocast structure and what this yaml is doing?

cmd/cedana-helper/install.sh
bsmithai: Yes we should, we are only tested and working on the linux 2 image in eks so I didn't add support for anything else. But I can change this.

api/services/cts.go
bsmithai: I have a retry on the helper

test.yaml
bsmithai: I docker compose mesocast but the nexrad 3 and nexrad 2 parsers are stateful and I wanted to test Cedana c/r of those containers specifically, since mesocast is being moved to EKS, I thought it be perfect for testing.

bsmithai: The mesocast parsers are like extreme for k8s testing though, which is good, they need minimum 12 cpus and 96 gb of ram, they on average use up 60gb

cmd/cedana-helper/install.sh
nravic: OK awesome, just toss a comment in here for now then

PR #145, opened 2024-03-13
Hotfix/arch

bsmithai:


cmd/cedana-helper/main.go
```diff
@@ -132,7 +117,7 @@ func runCommand(command string, args ...string) error {
 }
 
 func startDaemon() (int, error) {
-	cmd := exec.Command(daemonCommand, "daemon", "start&")
+	cmd := exec.Command("../../build-start-daemon.sh")
```
nravic, line 120: no issues w/ systemd here? happy to add a flag to the script that controls whether or not the daemon is started as a systemd daemon

nravic (approved):
small comment, lgtm otherwise

cmd/cedana-helper/main.go
bsmithai: No issues with running as a systemd daemon since this is technically on the ec2 instance so we get access to systemctl

PR #148, opened 2024-04-09
send gpuckpt and mem filepaths over RPC post gpu checkpoint

lianakoleva:
requires [cedana/cedana-gpu:feat/ced-263](https://github.com/cedana/cedana-gpu/tree/feat/ced-263)


bsmithai:
Are we wanting to unmarshal the paths from the checkpoint response and then doing something with them in this PR

bsmithai (approved):
lgtm

PR #158, opened 2024-04-29
run terminal checkpoint from job using gRPC

lianakoleva:
Requires https://github.com/cedana/cedana-benchmarking/tree/feat/ced-319

### DIFF
--- a/test/benchmarks/benchmark.py
+++ b/test/benchmarks/benchmark.py
@@ -343,35 +343,56 @@ def main(args):
         "'python3 1gb_pytorch.py'",
     ]
 
-    # run in a loop
-    num_samples = 5
-    for x in range(len(jobs)):
-        print("Starting benchmarks for job {} with command {}".format(jobs[x], cmds[x]))
-        job = jobs[x]
-        for y in range(num_samples):
-            jobID = job + "-" + str(y)
-            process_stats = run_exec(cmds[x], jobID)
-            # wait a few seconds for memory to allocate
-            time.sleep(5)
+    if "--correctness" in args:
+        job = "nn-1gb"
+        cmd = "'python3 1gb_pytorch_correctness.py'"
+        print("Starting correctness test for job {} with command {}".format(job, cmd))
+        jobID_base = job + "-base"
+        jobID_saved = job + "-saved"
+        process_stats_base = run_exec("'python3 1gb_pytorch_correctness.py nn-1gb-base'", jobID_base)
+        process_stats_saved = run_exec("'python3 1gb_pytorch_correctness.py nn-1gb-saved'", jobID_saved)
+        time.sleep(5)
+        run_checkpoint(daemon_pid, jobID_saved, output_dir, process_stats_saved)
+        time.sleep(3) # vary time and  check checkpoints
+        run_restore(daemon_pid, jobID_saved, output_dir)
+        time.sleep(3)
+        print("process_stats_base:\n", process_stats_base)
+        print("process_stats_saved:\n", process_stats_saved)
+        # do not terminate, wait for the jobs to exit and then compare checkpoints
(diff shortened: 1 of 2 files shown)


test/benchmarks/benchmark.py
```diff
@@ -343,35 +343,56 @@ def main(args):
...
+        cmd = "'python3 1gb_pytorch_correctness.py'"
+        print("Starting correctness test for job {} with command {}".format(job, cmd))
+        jobID_base = job + "-base"
+        jobID_saved = job + "-saved"
+        process_stats_base = run_exec("'python3 1gb_pytorch_correctness.py nn-1gb-base'", jobID_base)
+        process_stats_saved = run_exec("'python3 1gb_pytorch_correctness.py nn-1gb-saved'", jobID_saved)
+        time.sleep(5)
+        run_checkpoint(daemon_pid, jobID_saved, output_dir, process_stats_saved)
+        time.sleep(3) # vary time and  check checkpoints
+        run_restore(daemon_pid, jobID_saved, output_dir)
```
nravic, line 357: are you only running checkpoint/restore here to collect stats?

test/benchmarks/benchmark.py
```diff
@@ -343,35 +343,56 @@ def main(args):
...
-    if local:
-        return
-    # unique uuid for blob id
-    blob_id = "benchmark-data-" + str(time.time())
-    push_otel_to_bucket("/cedana/data.json", blob_id)
-    attach_bucket_id("benchmark_output.csv", blob_id)
-    push_to_bigquery()
-    # delete benchmarking folder
-    cleanup()
+    if "--correctness" in args:
```
nravic, line 346: I think we should eventually separate this from the benchmark.py script

test/benchmarks/benchmark.py
```diff
@@ -343,35 +343,56 @@ def main(args):
...
+        job = "nn-1gb"
+        cmd = "'python3 1gb_pytorch_correctness.py'"
+        print("Starting correctness test for job {} with command {}".format(job, cmd))
+        jobID_base = job + "-base"
+        jobID_saved = job + "-saved"
+        process_stats_base = run_exec("'python3 1gb_pytorch_correctness.py nn-1gb-base'", jobID_base)
+        process_stats_saved = run_exec("'python3 1gb_pytorch_correctness.py nn-1gb-saved'", jobID_saved)
+        time.sleep(5)
+        run_checkpoint(daemon_pid, jobID_saved, output_dir, process_stats_saved)
```
nravic, line 355: could you also update these functions to make the checkpoint/restore calls using grpc (instead of using the cli?)

test/benchmarks/benchmark.py
nravic: Lol kinda vague comment. Leaving a comment in here that we should separate it out is enough I think.

test/benchmarks/benchmark.py
lianakoleva: yes, i sometimes comment it out but i can make them more formal functions if you want

test/benchmarks/benchmark.py
lianakoleva: yes i was thinking that too -- one top-level script that calls either benchmarking or correctness, so all the CI infra is shared

test/benchmarks/benchmark.py
lianakoleva: sure, would you want me to update all of benchmark.py then for consistency?

nravic: yeah might as well! all of benchmark.py uses the same run_checkpoint, run_restore functions so just modifying those should be good

nravic (approved):

test/benchmarks/benchmark.py
nravic: can add this to the refactoring PR

PR #159, opened 2024-05-01
Simplify interface layer (CLI/API)

yashanand1910:
For details on all the changes, please check: https://linear.app/cedana/issue/CED-317/simplify-cedana-interface-layer-cliapi

About ~400 LOC were removed as part of this.

This shouldn't ideally break any functionality, however things have been moved around for better organization, and thus might break tests/CI.


cmd/dump.go
```diff
@@ -0,0 +1,269 @@
+package cmd
+
+// This file contains all the dump-related commands when starting `cedana dump ...`
+
+import (
+	"fmt"
+	"strconv"
+
+	"github.com/cedana/cedana/api"
```
nravic, line 9: unused import

api/process.go
```diff
@@ -0,0 +1,435 @@
...
+
+	state.JobState = task.JobState_JOB_RUNNING
+	state.PID = pid
+	if args.JID == "" {
+		state.JID = xid.New().String()
+	} else {
+		existingState, err := s.getState(args.JID)
+		if err != nil {
+			return nil, status.Error(codes.NotFound, "failed to check existing jobs")
+		}
```
nravic, line 57: This fails immediately when starting a process w/ exec in a clean setup (no /tmp/cedana.db file) with:
`2:18PM ERR start task failed error="rpc error: code = NotFound desc = failed to check existing jobs" file=/home/nravic/go/src/github.com/cedana/cedana/cmd/exec.go line=76`


My theory is nothing is being persisted to the db after the run (which is successful fwiw, I see a pid being spawned).

utils/logger.go
```diff
@@ -3,40 +3,51 @@ package utils
...
+	var output io.Writer = zerolog.ConsoleWriter{
+		Out: os.Stdout,
+	}
 
-	return log
+	logger = zerolog.New(output).
+		Level(logLevel).
+		With().
+		Timestamp().
+		Logger().Hook(LineInfoHook{})
```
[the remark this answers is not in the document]
yashanand1910: Oh, great! But we don't want line numbers for each message do we? It would clutter. Let me check this out, if we can configure it for only warning/errors and above levels.

api/process.go
yashanand1910: On it.

utils/logger.go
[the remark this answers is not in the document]
yashanand1910: Okay, `caller` is pretty concise. I changed it to use it when the currently set log level is debug or lower.

api/process.go
yashanand1910: Fixed. So any get call on a fresh DB was failing. Because put commands were the only ones doing the necessary bucket creation etc.

nravic:
@yashanand1910 if you resolve the conflicts I think this is ready to merge.

yashanand1910:
> @yashanand1910 if you resolve the conflicts I think this is ready to merge.

Merge whenever, all conflicts have been resolved.

nravic (approved):
send it!

PR #160, opened 2024-05-08
Refactor benchmarking + correctness

lianakoleva:


nravic (approved):
much cleaner, looks great!

PR #161, opened 2024-05-10
add crit show to cedana api

lianakoleva:


bsmithai (approved):
This lgtm. I still think it would be beneficial that we use their proto files and unmarshal the image files to go structs but not needed right now

PR #165, opened 2024-05-13
CED-233: GPU Checkpointing cleanup 

nravic:
Remove env var and trigger gpu checkpoints over rpc.


cmd/daemon.go
```diff
@@ -49,15 +42,15 @@ var startDaemonCmd = &cobra.Command{
 		if viper.GetBool("profiling_enabled") {
 			go startProfiler()
 		}
-
-		if viper.GetBool("gpu_enabled") {
-			err := pullGPUBinary(ctx, gpuControllerBinaryName, gpuControllerBinaryPath)
+		gpuEnabled, _ := cmd.Flags().GetBool(gpuEnabledFlag)
```
nravic, line 45: add flag to specify cuda path, we determine version

api/restore.go
```diff
@@ -516,6 +516,7 @@ func (s *service) restore(ctx context.Context, args *task.RestoreArgs) (*int32,
 	}
 
 	var gpuCmd *exec.Cmd
+	// No GPU flag passed in args - if state.GPUCheckpointed = true, always restore using gpu-controller
```
yashanand1910, line 519: Nice!

utils/config.go
```diff
@@ -17,6 +17,13 @@ const (
 	configFilePerm = 0644
 )
 
+const (
```
yashanand1910, line 20: I'm guessing this will be updated in a later PR?

yashanand1910 (approved):
Lgtm!

utils/config.go
nravic: what needs to be updated?

yashanand1910: Sorry, I meant the names of the GPU binaries/libs.

nravic: oh yeah we can change stuff around once we start pulling from cloudsmith

PR #166, opened 2024-05-14
CED-333: Stuffing more data into process state   

nravic:
Add more data to runc and process c/r response about machine state


bsmithai (approved):
lgtm

PR #167, opened 2024-05-14
Use process_json.diff_ckpts in correctness.py, add error msg

lianakoleva:
Example of diff checkpoint test at end of correctness execution:
Screenshot from 2024-05-14 16-17-05

Example of output when checkpoints are different:
Screenshot from 2024-05-14 16-07-50

Example of output when checkpoints are same:
Screenshot from 2024-05-14 16-00-55

### DIFF
--- a/test/benchmarks/correctness.py
+++ b/test/benchmarks/correctness.py
@@ -1,4 +1,5 @@
 import cedana_bindings as cedana
+from process_json import diff_ckpts, print_result
 import psutil
 import time
 from tplib import task_pb2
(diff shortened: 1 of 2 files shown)


nravic (commented):
Could you add a quick save_result function that saves the output of this to a csv or some sort of ordered output? That way we can run this at scale and process the data eleswhere too

lianakoleva:
Screenshot from 2024-05-15 12-08-34
Screenshot from 2024-05-15 12-07-42

nravic (approved):
lgtm!

PR #169, opened 2024-05-23
Correctness test cases

lianakoleva:
Default behavior shows test along with its status:
Screenshot from 2024-05-22 22-03-06
There are always two jobs running (job1 and job2). A terminal checkpoint is taken of each and used to determine if program behavior has been affected. Testcases that contain "c[x]" and "r[x]" mean that job[x] was checkpointed and restored, respectively. For instance, "c1c2r1r2" checkpoints job1, checkpoints job2, restores job1, and restores job2. 

Verbose mode shows complete output of each individual test:
Screenshot from 2024-05-22 22-05-02


lianakoleva:
depends on https://github.com/cedana/cedana-benchmarking/commit/e46de81d331ffc23df9dc8ed3e05754a3f03ac31

nravic (approved):
looks good to me!

PR #174, opened 2024-05-30
CED-365: Add CUDA version as option to pulling binary

nravic:


yashanand1910 (approved):
Looks good!

cmd/daemon.go
```diff
@@ -21,21 +23,27 @@ var daemonCmd = &cobra.Command{
 	Short: "Start daemon for cedana client. Must be run as root, needed for all other cedana functionality.",
 }
 
+var cudaVersions = map[string]string{
+	"cuda11_8": "cuda11_8",
+	"cuda12_1": "cuda12_1",
+	"cuda12_4": "cuda12_4",
+}
```
yashanand1910, line 30: Can make it such that user can pass flag like `--cuda 11.8`, which looks much cleaner than `--cuda cuda11_8`.
```suggestion
var cudaVersions = map[string]string{
	"11.8": "cuda11_8",
	"12.1": "cuda12_1",
	"12.4": "cuda12_4",
}
```

yashanand1910 (changes requested):
Minor UX comment.

yashanand1910 (approved):
Awesome! Good to merge now

PR #176, opened 2024-05-30
Watching restored PID in gofunc no longer limited to GPU-bound jobs

nravic:


api/restore.go
```diff
@@ -533,25 +533,28 @@ func (s *service) restore(ctx context.Context, args *task.RestoreArgs) (*int32,
...
+			if err != nil || !running {
+				break
 			}
+			// XXX YA: Sleeping on loop is always bad, should use a channel or cond var
+			time.Sleep(1 * time.Second)
+		}
+
+		if gpuCmd != nil {
 			s.logger.Debug().Msgf("process %d exited, killing gpu-controller", *pid)
 			gpuCmd.Process.Kill()
```
yashanand1910, line 553: Just a thought: what if there are other processes relying on GPU controller and we kill it. This would be an existing bug, not related to this PR, though.

yashanand1910 (approved):
Looks good! Left a comment.

api/restore.go
nravic: if it's multiple processes -> single GPU controller, yeah can see an issue

nravic: we could have a waitgroup or something in the case for multiple processes -> single GPU controller and wait for each process to die first, but imo getting ahead of ourselves on that

yashanand1910: Will create a separate ticket for this, you can merge this. One solution might be to simply kill it when the daemon is killed, as then it should not be used anyways.

PR #179, opened 2024-06-04
add regression test to check post-restore logging + fix stdout/stderr…

sanganip:
- add a regression test to check if the post-restore log file is not empty
- fix the process restore function to ensure that stdout and stderr are correctly redirected after restoring a process


nravic (approved):
great work!

PR #180, opened 2024-06-04
Add a PR template 

nravic:


yashanand1910:
Also, @nravic there's a way to enable automatic link resolution in GitHub for ticket numbers somewhere.
Such that, wherever we mention a ticket like `CED-XXX` it will be converted to a linear link.

nravic:
> Also, @nravic somewhere there's a way to enable automatic link resolution in GitHub for ticket numbers. Such that, wherever we mention a ticket like `CED-XXX` it will be converted to a linear link.

this is already set up!

PR #187, opened 2024-06-06
Use custom GPU binaries if custom path provided

yashanand1910:
## Describe your changes
If custom path for the GPU binaries is provided via the config items:

- `gpu_controller_path` (or env var `CEDANA_GPU_CONTROLLER_PATH`)

- `gpu_shared_lib_path` (or env var `CEDANA_GPU_SHARED_LIB_PATH`)

Then, it should not download these binaries into the default locations but use the binaries provided on these paths. This will help in testing/benchmarking.

## Issue ticket number 
CED-390


cmd/exec.go
```diff
@@ -59,7 +58,6 @@ var execTaskCmd = &cobra.Command{
 		gpuEnabled, _ := cmd.Flags().GetBool(gpuEnabledFlag)
 		if gpuEnabled {
 			logger.Info().Msgf("starting task w/ gpu enabled")
-			executable = fmt.Sprintf("LD_PRELOAD=%s %s", utils.GpuSharedLibPath, executable)
```
nravic, line 62: I think we should leave this in so customers don't have to manually LD_PRELOAD their gpu tasks. Does this not work on your machine @yashanand1910 ?

nravic (commented):
looks good! added a small comment

cmd/exec.go
yashanand1910: Oh! This was simply moved to the daemon. Check `process.go`, where we start the controller. Just thought to keep it at one place.

nravic: Ah missed that - thank you! Yeah better in process than cmd/exec.

nravic (approved):
lgtm!

PR #193, opened 2024-06-10
resolve containerd rootfs restores

bsmithai:
## Describe your changes
- Modified containerd checkpoint to use a custom config spec json opt
- Modify dumprootfs function to use this containerd checkpoint from `container/checkpoint.go`

## Issue ticket number 
[CED-395](https://linear.app/cedana/issue/CED-395/fix-rootfs-restore-invalid-media-type)

### DIFF
--- a/container/checkpoint.go
+++ b/container/checkpoint.go
@@ -1005,7 +1005,8 @@ func ContainerdCheckpoint(ctx context.Context, containerdClient *containerd.Clie
 	}
 
 	containerdContainer := &ContainerdContainer{
-		client: containerdClient,
+		Container: container,
+		client:    containerdClient,
 	}
 
 	task, err := container.Task(ctx, nil)
--- a/container/checkpoint.go
+++ b/container/checkpoint.go
@@ -985,7 +985,8 @@ type ContainerdContainer struct {
 	client *containerd.Client
 }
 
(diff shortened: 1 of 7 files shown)


container/checkpoint.go
```diff
@@ -1012,7 +983,189 @@ func CheckRuntime(current, expected string) bool {
 	return true
 }
 
-func runcCheckpointContainerd(ctx gocontext.Context, client *containerd.Client, task containerd.Task, opts ...CheckpointTaskOpts) (containerd.Image, error) {
+type ContainerdContainer struct {
+	containerd.Container
+	client *containerd.Client
+}
+
+func ContainerdCheckpoint(ctx context.Context, containerdClient *containerd.Client, id, ref string) error {
```
nravic, line 991: should we add a flag or input here to differentiate b/ween rootfs containerd c/r and just containerd c/r?

nravic (approved):
lgtm added a small comment

container/checkpoint.go
bsmithai: Honestly didn't add that because for full c/r I do not expect to use this outside of being used by orchestration. So a controller would c/r process state + rootfs. So I can rename this to a ContainerdRootfsCheckpoint or something similar

PR #197, opened 2024-06-11
Change default runc restore to not need existing container

yashanand1910:
## Describe your changes
This does not break any existing functionality, just adds a sane default.

This is to set a sane default when restoring a runc container using `cedana restore runc …`. Currently, it's not possible to restore a runc container without an existing running container that has a state.json and an init.pid in the bundle that's used to restore in an already created network namespace.

Now, by default, it should be possible to do a simple runc restore.

## Issue ticket number 
CED-405


yashanand1910:
WIP: adding new runc regression tests.

bsmithai (approved):
this lgtm, no breaking changes I don't think

nravic (approved):
lgtm! great work on the tests

test/regression/main.bats
```diff
@@ -70,3 +70,60 @@ load helper.bash
...
+  run mkdir -p $bundle/rootfs
+  run tar -C $bundle/rootfs -xzf alpine-minirootfs-3.10.1-x86_64.tar.gz
+  sudo chown -R root:root $bundle
+
+  # create a runc container
+  runc_exec $bundle $job_id
+  run sleep 1
+
+  # check if container running correctly, count lines in output file
+  [[ -f $out_file]]
```
nravic, line 92: $out_file ]], not $out_file]]

yashanand1910:
@nravic The tests work now. Apparently, there is an open CRIU issue that's mentioned in a runc ticket (https://github.com/opencontainers/runc/issues/1218) that was causing it. Dumping runc containers that have `terminal: false` in `config.json` does not work? So I set that and I'm now providing a `--console-socket` during launch as well as restore.

bsmithai:
> @nravic The tests work now. Apparently, there is an open CRIU issue that's mentioned in a runc ticket ([opencontainers/runc#1218](https://github.com/opencontainers/runc/issues/1218)) that was causing it. Dumping runc containers that have `terminal: false` in `config.json` does not work? So I set that and I'm now providing a `--console-socket` during launch as well as restore.

Oh yeah I remember this! Good find

PR #200, opened 2024-06-14
Fix exit status on errors

yashanand1910:
## Describe your changes
Currently, if the daemon or any other command in `cedana [cmd] …` results in an error, it does not exit with a non-zero code. This creates problems for the benchmark and tests that rely on error code to check for correct execution.

Also, fixes the issue where usage is printed even if it's not a usage error.

## Issue ticket number 
CED-415


nravic (approved):
tested locally with some bad commands and a regular checkpoint/restore. lgtm!

PR #212, opened 2024-06-25
Fix: Use benchmarking scripts instead of docker container in CI

nravic:
## Describe your changes
We were previously using the docker container to run checkpoint/restore benchmarks, but it's cleaner to just use the host itself, and potentially (later) checkpoint/restore containers in the benchmarks.


yashanand1910 (approved):
Lgtm!

PR #230, opened 2024-07-10
Feat: Add healthcheck grpc endpoint (& detailed healthcheck)

nravic:


.gitmodules
yashanand1910, line 1: Guessing this was temporary?

api/server.go
```diff
@@ -286,3 +295,85 @@ func loggingUnaryInterceptor(logger *zerolog.Logger) grpc.UnaryServerInterceptor
...
+	}
+
+	defer cmd.Process.Kill()
+
+	var opts []grpc.DialOption
+	opts = append(opts, grpc.WithTransportCredentials(insecure.NewCredentials()))
+
+	gpuConn, err := grpc.Dial("127.0.0.1:50051", opts...)
+	if err != nil {
+		return err
```
yashanand1910, line 358: Same as above. Let's use error return only if the health check itself has a problem.

api/server.go
```diff
@@ -286,3 +295,85 @@ func loggingUnaryInterceptor(logger *zerolog.Logger) grpc.UnaryServerInterceptor
...
+	}
+
+	defer gpuConn.Close()
+
+	gpuServiceConn := gpu.NewCedanaGPUClient(gpuConn)
+
+	args := gpu.HealthCheckRequest{}
+	gpuResp, err := gpuServiceConn.HealthCheck(ctx, &args)
+	if err != nil {
+		return err
```
yashanand1910, line 368: Same as above.

api/server.go
```diff
@@ -286,3 +295,85 @@ func loggingUnaryInterceptor(logger *zerolog.Logger) grpc.UnaryServerInterceptor
...
+	}
+
+	if !gpuResp.Success {
+		return fmt.Errorf("GPU health check failed")
```
yashanand1910, line 372: Should add reason to response instead. Same as above.

api/services/cts.go
```diff
@@ -44,6 +45,42 @@ func (c *ServiceClient) Close() {
...
+	} else {
+		return false, nil
+	}
+}
+
+func (c *ServiceClient) DetailedHealthCheck(ctx context.Context, args *task.DetailedHealthCheckRequest) (*task.DetailedHealthCheckResponse, error) {
+	ctx, cancel := context.WithTimeout(ctx, DEFAULT_PROCESS_DEADLINE)
+	defer cancel()
+	resp, err := c.taskService.DetailedHealthCheck(ctx, args)
+	if err != nil {
```
yashanand1910, line 77: Should use `err` only if the health check itself failed due to an internal error. Otherwise, should check if the `UnhealthyReasons` list is empty and respond based on that.

yashanand1910 (changes requested):
Left a few comments on error handling. I believe it's better to return an `error` if the health check itself failed due to internal reasons. Whereas, if there are failures in the 'health checklist' they are added to `UnhealthyReasons`. Can discuss offline to make this consistent.

.gitmodules
```diff
@@ -1,3 +1,4 @@
 [submodule "api/services/cedana-api"]
 	path = api/services/cedana-api
 	url = https://github.com/cedana/cedana-api
+	branch = feat/ced-457
```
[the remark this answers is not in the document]
nravic: yep!

yashanand1910 (approved):
Looks great!

PR #232, opened 2024-07-11
feat: add arm64 support & bash script fixes

swarnimarun:
## Describe your changes

CRIU supports arm64 by default, so we should be without issues be able to support it as well

Also fixes some issues with script, such as adding `set -e` and removing hard-coded use of `sudo`.

## Issue ticket number 

Fixes CED-507

### Have I performed a self-review?

Yes

### Have I added regression or unit tests (where it makes sense) for my changes?

No, none needed. It's tested by smoke and regression tests.

### Have I updated the README if changes have resulted in it being out of date?

Documentation update is required.

### Have I checked to ensure my PR only introduces the changes it's intended to? E.g. No extraneous formatting changes.

There are a couple of minor changes to ensure everything compiles correctly. There was an issue with tests which broke my local builds for tests.

### Is this a breaking change?

Yes, we use `set -e`, which means our build daemon bash script can fail if any of the commands fail. Originally this wouldn't happen but I believe this should be the expected behaviour so that we can report and fix it.


.github/workflows/release.yml
```diff
@@ -31,17 +41,16 @@ concurrency:
   cancel-in-progress: ${{ github.ref != 'refs/heads/cedana-dev' }}
 
 jobs:
-  release:
+  release_x86:
```
swarnimarun, line 44:
```suggestion
  release:
```

nravic (approved):
lgtm!

PR #234, opened 2024-07-11
Fix/ced 484

bsmithai:
## Describe your changes
- Modify changed files in CRIO rootfs checkpoint to be under root ownership

## Issue ticket number
[CED-484](https://linear.app/cedana/issue/CED-484/sysbox-rootfs-file-perm-issues)


utils/logger.go
```diff
@@ -15,6 +17,22 @@ const (
 	LOG_CALLER_SKIP      = 3 // stack frame depth
 )
 
+type contextKey string
+
+const loggerKey = contextKey("logger")
```
nravic, line 22: not that this isn't a bad thing but why introduce it with this PR? the place you use it (api/crio.go) still has access to s.logger. Does ctx need a logger for some other reason?

nravic (changes requested):
Let's also rename crioPush in cmd while we're at it

utils/logger.go
bsmithai: I have it being picked up in api/crio/crio.go through ctx, I didn't want to add it as another argument

nravic (approved):
lgtm!

PR #236, opened 2024-07-14
Make cedana exec fully manage process + GPU controller

yashanand1910:
## Describe your changes
Currently, if the daemon exits, the process is allowed to run on it's own. For simple, CPU-only jobs, there's no value of using `cedana exec` if it's not managing its lifecycle, because you can always C/R an unmanaged process as well (process not started by cedana exec).

In the case of GPU-enabled, `cedana exec` spawns the GPU controller which will continue to run without ever getting killed if the daemon exits prematurely. The lifecycle of the GPU controller must be tied to the process using it. As the GPU controller is only ever spawned for a manged `cedana exec -g` process, this becomes simple as the process as well as the GPU controller can be killed if the daemon exits.

Change highlights:
- Local DB is only used for managed jobs now, if starting through `cedana exec`. If `cedana dump/restore process` is used, the DB won't be touched. Same goes for any other form of unmanaged dump/restore.
- If the daemon exits prematurely, all managed jobs will be killed and their DB entries will be updated gracefully before it exits. This includes any _restored_ managed jobs, as well as any instances of running GPU controller.
- Fix bug where daemon interruption would only wait for RPC calls to finish, but not any of the background go routines.

To clarify, by 'unmanaged' I mean any process/container whose lifecycle is not tied with the daemon.

## Issue ticket number 
CED-504

> Added 2 tests for checking managed job graceful exit before and after C/R.
> Includes some minor changes to log formatting for convenience
> Nope.


nravic (approved):
looks great!

PR #242, opened 2024-07-19
Kata container integration

sanganip:
## Issue ticket number 
CED-506


.gitmodules
```diff
@@ -1,3 +1,4 @@
 [submodule "api/services/cedana-api"]
 	path = api/services/cedana-api
 	url = https://github.com/cedana/cedana-api
+	branch = feat/CED-506
```
yashanand1910, line 4: Leaving a comment here as a reminder to revert this.

api/kata.go
```diff
@@ -0,0 +1,103 @@
...
+			Message:      fmt.Sprintf("Dumped process %d to %s", pid, args.Dir),
+			CheckpointID: state.CheckpointPath, // XXX: Just return path for ID for now
+		}
+	}
+
+	err = s.updateState(ctx, state.JID, state)
+	if err != nil {
+		st := status.New(codes.Internal, fmt.Sprintf("failed to update state with error: %s", err.Error()))
+		return nil, st.Err()
+	}
```
yashanand1910, line 63: This can be removed. Saving state to local DB is now only done for managed jobs, ie. jobs started through `cedana exec`. See #236.

api/kata.go
```diff
@@ -0,0 +1,103 @@
...
+	return &resp, err
+}
+
+
+
+//////////////////////////
+///// Kata Utils //////
+//////////////////////////
+
+func childPidFromPPid(ppid int32) (int32) {
```
yashanand1910, line 76: This could just return an error as well `(int32, error)` so we don't have to proceed with PID -1 where it's called.

api/server.go
```diff
@@ -42,6 +44,10 @@ const (
 	GPU_CONTROLLER_LOG_PATH = "/tmp/cedana-gpucontroller.log"
 )
 
+const (
+	port = 9999
+)
```
yashanand1910, line 49: Can rename this variable as `PORT` or `VSOCK_PORT` as all our constants are in uppercase.

api/server.go
```diff
@@ -62,7 +68,7 @@ type Server struct {
 	listener   net.Listener
 }
 
-func NewServer(ctx context.Context) (*Server, error) {
+func NewServer(ctx context.Context, kataEnabledFlag bool) (*Server, error) {
```
yashanand1910, line 71: Would be nice to create a `ServerOptions` struct that is instead passed to the `NewServer` function. This would allow for more flexibility in the future. For example, if we wanted to add a `MaxConnections` field to the `ServerOptions` struct, we could do so without breaking the signature.

Also, it would be better to name the flag `vsockEnabled` instead as there's nothing Kata specific about it. In future, we might want to add support for other VM/runtimes that use vsock.

api/services/cts.go
```diff
@@ -19,6 +24,11 @@ const (
 	DEFAULT_RUNC_DEADLINE       = 10 * time.Minute
 )
 
+const (
+	port = 9999
+)
```
yashanand1910, line 29: Should simply use global variable declared in `server.go`.

api/services/cts.go
```diff
@@ -41,6 +51,32 @@ func NewClient() (*ServiceClient, error) {
 	return client, err
 }
 
+func NewKataClient(vm string) (*ServiceClient, error) {
```
yashanand1910, line 54: Should instead call this `NewVsockClient` as there's nothing Kata-specific about it.

cmd/daemon.go
```diff
@@ -165,6 +167,7 @@ func init() {
 	daemonCmd.AddCommand(startDaemonCmd)
 	daemonCmd.AddCommand(checkDaemonCmd)
 	startDaemonCmd.Flags().BoolP(gpuEnabledFlag, "g", false, "start daemon with GPU support")
+	startDaemonCmd.Flags().BoolVarP(&kataEnabledFlag, "kata", "k", false, "start daemon inside Kata VM")
```
yashanand1910, line 170: Why are we using `BoolVarP` instead of `BoolP`? We could just retrieve the value just like other flags.

utils/config.go
```diff
@@ -48,12 +48,13 @@ type (
...
-	if err != nil {
-		return err
-	}
+	// user, err := getUser()
+	// if err != nil {
+	// 	return err
+	// }
 
-	homeDir := user.HomeDir
+	// homeDir := user.HomeDir
```
yashanand1910, line 56: Leaving this comment as a reminder to revert/find a fix for this.

yashanand1910 (changes requested):
Awesome work! I know this is still a work in progress, but I have left a few preliminary comments on code quality. I'll review it again once it's finalized.

utils/config.go
yashanand1910: @sanganip Can use this to specify `--config-dir` https://github.com/cedana/cedana/pull/256 once merged.

api/restore.go
```diff
@@ -84,30 +85,32 @@ func (s *service) prepareRestore(ctx context.Context, opts *rpc.CriuOpts, args *
...
-	if err != nil {
-		s.logger.Error().Err(err).Msg("error creating logfile")
-		return nil, nil, nil, err
-	}
-
-	for _, f := range open_fds {
-		if strings.Contains(f.Path, "pts") {
-			isShellJob = true
+		// create logfile for redirection
+		filename := fmt.Sprintf(RESTORE_OUTPUT_LOG_PATH, fmt.Sprint(time.Now().Unix()))
```
yashanand1910, line 92: Do we not need to open the `KATA_RESTORE_OUTPUT_LOG_PATH` here?

api/restore.go
```diff
@@ -603,6 +606,57 @@ func (s *service) restore(ctx context.Context, args *task.RestoreArgs) (*int32,
...
+			// strip leading slash from f
+			f.Path = strings.TrimPrefix(f.Path, "/")
+
+			extraFiles = append(extraFiles, file)
+			opts.InheritFd = append(opts.InheritFd, &rpc.InheritFd{
+				Fd:  proto.Int32(2 + int32(len(extraFiles))),
+				Key: proto.String(f.Path),
+			})
+		}
+	}
```
yashanand1910, line 648: Could just add it to `prepareRestore` as most of it's duplicate code.

yashanand1910 (changes requested):
Looked at the new changes, left a few more comments.

cmd/daemon.go
sanganip: BoolP kept giving me compilation issues. I guess there can only be 1 BoolP flag?

yashanand1910 (approved):
Great!!

PR #251, opened 2024-07-24
fix: symlink ownership changes

bsmithai:
## Describe your changes
- Lchown files that are of mode symlink

## Issue ticket number
[CED-538](https://linear.app/cedana/issue/CED-538/fix-symlink-ownership-not-changing-in-crio-sysbox-rootfs)

### DIFF
--- a/api/crio/crio.go
+++ b/api/crio/crio.go
@@ -13,6 +13,7 @@ import (
 	"path/filepath"
 	"regexp"
 	"strings"
+	"syscall"
 
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/session"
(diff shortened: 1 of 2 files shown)


api/crio/crio.go
```diff
@@ -227,8 +232,25 @@ func RootfsCheckpoint(ctx context.Context, ctrDir, dest, ctrID string, specgen *
 
 	for _, change := range rootFsChanges {
 		fullPath := filepath.Join(tmpRootfsChangesDir, change.Path)
-		if err := os.Chown(fullPath, 0, 0); err != nil {
-			fmt.Printf("failed to change ownership for %s: %s", fullPath, err)
+
```
nravic, line 235: I think a nice big comment here explaining what this is and why we're doing it here would be swell

nravic (changes requested):

api/crio/crio.go
bsmithai: done

nravic (approved):
lgtm

PR #252, opened 2024-07-25
fix: support auto fetch root user

swarnimarun:
## Describe your changes

Currently, with root user, if env variable is not set we fail.
This means most non-sudo envs fail to get user.


nravic (approved):
looks good!

PR #256, opened 2024-07-28
Add `--config` and `--config-dir` flags to daemon

yashanand1910:
## Describe your changes
The config is now created only when running `cedana daemon start …`. And these flags are added to this command. This is because the daemon does all the actual work and the config should be tied to it.

These CLI options will allow specifying a custom configuration to the cedana daemon.

- `--config`: Specify a custom config JSON string that will be used to override config. If there's an existing config, this will be merged with that. If there's no existing config, this will be merged with the default config, and used as the config just this once (not saved).

- `--config-dir`: Specify the location for storing config file. Any existing config file at this
  location will be discovered, or this will be used to create a new default config (or if config was
  provided via `--config`).

## Issue ticket number 
CED-541

> Should add a usage section in our README for all the flags through another ticket.
> This could be a breaking change, since the config is now only created for the first time when
> running `cedana daemon start …`.


nravic (commented):
Looks good! Re: testing - how do you feel about a command that returns the config that cedana has? It's a good spot validation that can be used in prod when debugging to look at what cedana "sees" as config. 

A test could then be creating a json file as config, passing it to cedana via --config, and then validating output

nravic (approved):
lgtm!

PR #263, opened 2024-08-01
fix: install packages all at once

swarnimarun:
## Describe your changes

Should improve install performance.

Note:
Tested on M1 Mac, for reinstallation, the performance goes from 10s to under 1sec.


swarnimarun:
Removing go plus, this should improve our first install times.

Also added a SKIP SETUP flag to skip setup entirely. Will make a separate PR to track and pass the flag to the env.

cmd/scripts/k8s/setup-host.sh
```diff
@@ -6,28 +6,26 @@ cp /usr/local/bin/build-start-daemon.sh /host/build-start-daemon.sh
...
     apt-get update
 
-    apt-get install -o Dpkg::Options::="--force-confdef" -o Dpkg::Options::="--force-confold" golang-github-containers-image golang-github-containers-common
+    apt-get install -y -o Dpkg::Options::="--force-confdef" -o Dpkg::Options::="--force-confold" golang-github-containers-image golang-github-containers-common
 
-    for pkg in $APT_PACKAGES; do
-        apt-get install -y $pkg || echo "Failed to install $pkg"
-    done
+    # install all packages at once
+    apt-get install -y "${APT_PACKAGES[@]}" || echo "Failed to install $pkg"
```
nravic, line 24: my only beef with this is that previously, by iterating through packages, if a package didn't exist we would continue and install all the rest (instead of failing). Now that packages are trimmed in size though maybe it's not such a bad thing bc these are all necessary.

nravic (approved):

PR #265, opened 2024-08-02
feat: faster cedanas restarts with custom cli flag

swarnimarun:
### Describe your changes

skips setup.


nravic (approved):
looks good, makes sense!

PR #272, opened 2024-08-04
Fix `--version` flag for cedana

yashanand1910:
## Describe your changes
It makes sense to embed the version from `git describe --tags` whenever building cedana. This way whenever we manually tag on GitHub, the release CI would automatically get that.

There were some simplifications to the CI workflows and the Dockerfile as a result, so on next release we should we keep an eye for any errors.

## Issue ticket number 
CED-578 & CED-586


.github/workflows/release.yml
```diff
@@ -51,6 +51,8 @@ jobs:
         with:
           fetch-depth: 0
           submodules: recursive
+          fetch-tags: true
```
swarnimarun, line 54: I believe fetch-depth 0 should fetch the entire history. Also note, fetch-depth 0 fetch the origin main head, which is fine for release but I am not sure if it's the correct behaviour we want.

swarnimarun (changes requested):

.github/workflows/release.yml
```diff
@@ -84,6 +86,7 @@ jobs:
         uses: actions/checkout@v4
         with:
           submodules: recursive
+          fetch-tags: true
```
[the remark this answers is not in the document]
yashanand1910: Good callout, as I'm reading online this should work when the action triggered by a tag. Still, will test it out!

.github/workflows/release.yml
yashanand1910: Yeah, I wasn't sure why this was `0` in the first place, so I left it. It should be `1` here too. @nravic Was there a reason for this being `0`?

nravic: Hmm no reason

.github/workflows/release.yml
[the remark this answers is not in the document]
yashanand1910: @swarnimarun Added a regression test to see if `cedana --version` works correctly. I think this should be sufficient to see if the `fetch-tags` is working correcly or now.

nravic (approved):
looks good!

PR #273, opened 2024-08-04
Read viper config only inside daemon

yashanand1910:
## Describe your changes
Currently the config is read at places in `cmd`. Since the config is now tied to the daemon, they should only be read inside the daemon.

This makes sense otherwise we would have different behaviour when accessing through CLI vs. directly through gRPC.

Also moves the `Config` type to `types` package as this may be commonly imported by external modules (e.g. Beta9).

## Issue ticket number 
CED-577

> Potentially changes the behaviour of remote checkpoint/restores, as now the env var `CEDANA_REMOTE` must be set for the daemon for it to look for remote checkpoints. Not sure if this is used rn, but should probably be changed to a CLI flag for the C/R commands anyway.


nravic (approved):
makes a ton of sense, good work

nravic (approved):
looks good!

PR #276, opened 2024-08-05
Revert "update: match bin commit sha in image"

swarnimarun:
Just checked, we now plan to use `git describe --tags` which doesn't add the commit hash to version field. If that is merge, it will fail all release actions after they have successfully pushed the images.

The relevant PR #272 

Not a breaking change.


yashanand1910:
@swarnimarun Can cancel this, I already took care of it in #272.

PR #284, opened 2024-08-09
Fix binary not built for lower glibc versions

yashanand1910:
## Describe your changes
Everywhere in the CI, where we are building the cedana binary, the glibc version is 2.36. To support even just the base image ubuntu 20.04, we need to link with glibc version that is at most 2.30 (since it's forward compatible).

## Issue ticket number 
CED-586

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,24 +1,29 @@
-FROM golang:1.22-bookworm as builder
+FROM golang:1.22-bullseye as builder
 
 WORKDIR /app
 ADD . /app
 
-RUN apt-get update \
-    && DEBIAN_FRONTEND=noninteractive \
-    apt-get install --no-install-recommends --assume-yes \
-    wget git make curl libnl-3-dev libnet-dev libbsd-dev runc libcap-dev \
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -1,24 +1,29 @@
...
-    && DEBIAN_FRONTEND=noninteractive \
-    apt-get install --no-install-recommends --assume-yes \
-    wget git make curl libnl-3-dev libnet-dev libbsd-dev runc libcap-dev \
-    libgpgme-dev btrfs-progs libbtrfs-dev libseccomp-dev libapparmor-dev libprotobuf-dev \
-    libprotobuf-c-dev protobuf-c-compiler protobuf-compiler python3-protobuf software-properties-common zip
+RUN <<EOT
+set -eux
+DEBIAN_FRONTEND=noninteractive
+apt-get update
+apt-get install -y wget git make curl libnl-3-dev libnet-dev libbsd-dev runc libcap-dev
```
nravic, line 10: Why do this instead of how we were installing before?

yashanand1910: Just cleaner, plus allows us to `set -eux`, so on failure during build, we get better message.

yashanand1910: Thought it's quite tedious to keep using `&&` you know.

PR #287, opened 2024-08-09
Add a CLI config section

yashanand1910:
## Describe your changes
As of this PR: https://github.com/cedana/cedana/pull/256, the config file is now only for the daemon. As the CLI is now independent of daemon stuff, we still need a way to configure CLI options. 

This PR:
- Adds a new section called `cli` in `client_config.json`.
- Loads config during start itself now, instead of when daemon starts, as CLI options need to be available for other CLI commands.
- The `--config` & `--config-dir` options are now flags available to the `cedana ...` cmd instead of `cedana daemon start`.

## Issue ticket number 
CED-595

Does not touch existing daemon config at all, so no.

### DIFF
--- a/api/services/cts.go
+++ b/api/services/cts.go
@@ -284,7 +284,7 @@ func (c *ServiceClient) GetConfig(ctx context.Context, args *task.GetConfigReque
 
 func getDefaultCallOptions() []grpc.CallOption {
 	opts := []grpc.CallOption{}
-	if viper.GetBool("wait_for_ready") {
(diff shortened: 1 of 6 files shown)


yashanand1910:
@nravic Updating the PR, to do what you had suggested earlier. Simply have a `cli` section in `client_config.json` instead. Where the daemon can simply use options that are meant for it. Having separate config file for CLI would have also required us to add a separate `--config` & `--config-dir` just for CLI, which would have made it more complex.

nravic (approved):
lgtm!

PR #288, opened 2024-08-09
Add `wait_for_ready` option to CLI config

yashanand1910:
## Describe your changes
This would allow cedana commands to be executed while the daemon is not up yet, or has possibly crashed and is restarting. Could be useful in many cases, but currently need it for the design we're using for the Beam integration.

Please also look at #287, as this is based on that. Can be merged in any order, though.

## Issue ticket number 
CED-594

> Nope. This will turned off by default.


nravic (approved):
makes sense to me!

nravic (approved):
lgtm!

PR #290, opened 2024-08-12
fix: always have skip in flight true

bsmithai:


nravic (changes requested):
I would add a truncated version of the comment from the ticket: 
> Currently CRIU will fail if we try to checkpoint a container that has tcp connections that are not fully established. Currently there is a [PR](https://github.com/cedana/cedana/pull/290) that has a quick fix which just introduces the skip in-flight connections. Ideally we wait to checkpoint until all tcp connections are in established states and we allow for setting a timeout on waiting for that in checkpoints. If we exceed the timeout, then we default to checkpointing with skip in-flight opt.

also we should talk about what implementing a barrier like this might look like

PR #291, opened 2024-08-13
Add `--attach` flag to cedana exec

yashanand1910:
## Describe your changes
Similar to `docker start --attach`, it makes sense to have a flag that attaches the process stdout/stderr. This makes sense when we need `cedana exec` to execute a process with it's lifecycle and exit code tied to it. 

There is also minimal stdin and signal handling. Standard interrupts that will terminate a process will work, but are not forwarded exactly. In future, we could forward the exact interrupts to the process.

This is needed for the current design of the Beam integration, where we are simply replacing the entrypoint of the container.

Also helpful for a future simplification to GPU benchmarks, as currently the benchmark waits for a command to exit and reads its stdout/stderr to proceed.

## Issue ticket number 
CED-589

> Added a new test for this, checking for exit code and stdout.
> Breaks no existing functionality. This flag is off by default.


nravic (approved):
regression tests are a beaut

PR #293, opened 2024-08-13
hotfix: Don't error if GPU restore stats is nil

yashanand1910:
## Describe your changes
This is needed for better backward compatibility, as it's older versions of `cedana-gpu` to always fail on `cedana restore`.

## Issue ticket number 
NA

### DIFF
--- a/api/restore.go
+++ b/api/restore.go
@@ -666,12 +666,11 @@ func (s *service) gpuRestore(ctx context.Context, dir string, uid, gid int32, gr
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


nravic (commented):
why would GPU controller not return stats though?

yashanand1910:
It does not on the main branch. We don't want `cedana dump` to fail with older versions of controller randomly.

yashanand1910:
Plus, stats being nil, shouldn't be a restore failure. It's too strict.

nravic (approved):

PR #297, opened 2024-08-14
Add `--attach` flag for `cedana restore job`

yashanand1910:
## Describe your changes
image

API PR: https://github.com/cedana/cedana-api/pull/17

Just like for `cedana exec`, this flag will cause the restore command to be attached to the restored processes stdin/stdout/stderr, and eventually exit with its exit code.

Creates temporary pipes, that are then inherited by the restored process replacing stdout/stdin/stderr. The daemon then simply reads/writes to these pipes and streams the data to/from the client.

*Additional change*: Managed jobs (started from `cedana exec`), even though run in their own session, are still children of the daemon, allowing us to `wait` on them. When restoring these jobs, they used to be no longer a child of the daemon, and the daemon could not wait on them. This was changed now so that whenever we are restoring a managed job, the CRIU option `--restore-sibiling` is used instead (see https://criu.org/Tree_after_restore). Note that the daemon always needs to wait on managed jobs to do cleanup - earlier this was done by spinning and checking process state which was painfully slow.

## Issue ticket number 
CED-612

> Added.
> See 'additional change' above.


nravic (approved):
lgtm

PR #300, opened 2024-08-15
Capture more info during release & also post on slack

yashanand1910:
## Describe your changes
Would make sense to capture binary size changes from last, at least.

On release, a notification on slack using our CI bot would be nice so we are sure to keep an eye. Drafts a custom Slack message using the [Block Kit](https://api.slack.com/block-kit), so we can modify it as per need. For e.g., we could add a deps diff later on. 

Similar script is used for GPU benchmark Slack summaries as well, so it would make sense to create our own github action at some point, so it can be easily added to all repos. No action out there provides all the exact stuff we need.

## Issue ticket number 
CED-592

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -365,5 +365,5 @@ jobs:
           SLACK_WEBHOOK_TYPE: INCOMING_WEBHOOK
         with:
           payload: |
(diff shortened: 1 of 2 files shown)


nravic (approved):
looks good!

PR #301, opened 2024-08-15
Fix job state update on early exit

yashanand1910:
## Describe your changes
There's a race condition where the daemon does not update the job state, if the program exits too early:

```bash
10:57AM INF process exited PID=29533 status=127
10:57AM INF managing process JID=ced PID=29533
10:57AM DBG gRPC request succeeded method
```
## Issue ticket number 
CED-620

> Simplifies the updation of state for both after `exec`ing and `restore`ing the job, so that the state is updated consistently. Also, ensures that the state is only updated in `process.go` and not touched by any of the internal functions.

### DIFF
--- a/api/process.go
+++ b/api/process.go
@@ -213,10 +213,8 @@ func (s *service) startHelper(ctx context.Context, args *task.StartArgs, stream 
 
 	state := &task.ProcessState{}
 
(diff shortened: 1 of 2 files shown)


lianakoleva (approved):
lgtm, thx

PR #304, opened 2024-08-16
feat: add job service API

swarnimarun:
### Describe your changes

Adds another service called job service for managing external API interface to daemon to manage checkpoint restores as jobs and job runners.

Uses sqlite for just metadata storage.

### Not part of this PR. This will make the PR too big.
- Add support for cancellation, perhaps a cancellation token.

### Issue ticket number

Fixes CED-599, CED-600

### Checklist before requesting a review


yashanand1910:
@swarnimarun The change of directory structure makes, but would suggest making such change as a quick separate PR, as this might be a long running PR and you will likely get terrible merge conflicts keeping in sync with main.

pkg/jobservice/service.go
```diff
@@ -0,0 +1,297 @@
...
+	db.SetMaxOpenConns(1)
+	db.SetMaxIdleConns(1)
+
+	if _, err := db.ExecContext(context.Background(), schemaSetup); err != nil {
+		return nil, err
+	}
+
+	config, err := detectConfig()
+	if err != nil {
+		// couldn't figure out runtime
```
nravic, line 42: worth creating an error here that explicitly says so because detectConfig just returns nil

nravic (approved):
looks good! small comment. I'd want to see tests for this, but that can be a separate PR like we discussed

PR #309, opened 2024-08-22
dump and restore --stream integration

lianakoleva:
## Describe your changes

Fully integrate image streaming with compression into cedana for straightforward use.

Example usage:
ced_stream_pytorch
ced_stream_test

Dependencies: 
- https://github.com/cedana/cedana-image-streamer
- https://github.com/lianakoleva/criu

## Issue ticket number
CED-520 

- [X] Have I performed a self-review?
Yes, I ensured that standard process C/R is not affected:
ced_standard_pytorch
ced_standard_test
- [X] Have I updated the README if changes have resulted in it being out of date?
Nothing is necessarily out of date, and I don't think usage is changed significantly enough to warrant its own README page in the docs. I am open to adding it in somewhere, but not sure where would be a good spot.
- [X] Have I checked to ensure my PR only introduces the changes it's intended to? E.g. No extraneous formatting changes.
Yes.
- [X] Is this a breaking change? If yes, please describe areas affected and request reviews from developer stakeholders. 
No, functionality stays the same when stream flag is omitted.


pkg/api/dump.go
```diff
@@ -254,25 +257,47 @@ func (s *service) runcDump(ctx context.Context, root, containerID string, pid in
...
 	// so have to ensure that image files aren't tampered with
-	s.postDump(ctx, imagePath, state, false)
+	s.postDump(ctx, imagePath, state, false, nil)
 
 	return nil
 }
 
+func (s *service) setupStreamerCapture(dumpdir string) *exec.Cmd {
+	buf := new(bytes.Buffer)
+	cmd := exec.Command("sudo", "cedana-image-streamer", "--dir", dumpdir, "capture")
```
nravic, line 275: this fails if cedana-image-streamer isn't in $PATH. I would check if it exists at the cmd level (where --stream is passed) and not progress if so

pkg/api/dump.go
```diff
@@ -129,14 +131,13 @@ func (s *service) prepareDump(ctx context.Context, state *task.ProcessState, arg
...
 	}
-
 	elapsed := time.Since(start)
 	stats.PrepareDuration = elapsed.Milliseconds()
 
 	return dumpDirPath, nil
 }
 
-func (s *service) postDump(ctx context.Context, dumpdir string, state *task.ProcessState, stream bool) {
+func (s *service) postDump(ctx context.Context, dumpdir string, state *task.ProcessState, stream bool, cmd *exec.Cmd) {
```
nravic, line 140: I think you can get rid of an argument here and not pass the stream bool, and maybe rename cmd to streamCmd? Then just check if streamCmd is nil

nravic (changes requested):
lgtm! small changes

nravic (approved):
lgtm! great work

PR #313, opened 2024-08-28
Fix GPU controller error handling and log dump

yashanand1910:
## Describe your changes
Currently, if the GPU controller exits prematurely, the program will keep waiting indefinitely. The
program should be killed. Same behaviour is needed for after a GPU process is restored.

Also, logs after GPU controller exit were not being captured. They should be dumped in the daemon
logs. This was not happening because the stdout/stderr was being set after starting the process.

## Issue ticket number
CED-631

> Nope.


nravic (approved):
perfect

PR #315, opened 2024-09-03
Add `--leave-running` flag for runc

yashanand1910:
## Describe your changes
Currently, a default value (true) is used for this CRIU option. We need an option to override from CLI.

BREAKING: The new default will be, thus, false. And this flag must be explicitly passed.

## Issue ticket number
CED-639

> The new default will be, thus, false. And this flag must be explicitly passed.

### DIFF
--- a/cmd/dump.go
+++ b/cmd/dump.go
@@ -4,6 +4,8 @@ package cmd
 
 import (
 	"fmt"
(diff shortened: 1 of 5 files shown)


nravic (approved):
lgtm

PR #320, opened 2024-09-03
Hotfix: nil dereference on process state

yashanand1910:
## Describe your changes
State returned by `generateState` could be `nil` without returning error.

## Issue ticket number
NA


pkg/api/state.go
```diff
@@ -173,37 +173,43 @@ func (s *service) generateState(ctx context.Context, pid int32) (*task.ProcessSt
...
-	vcpus, _ := cpu.Counts(true)
-	state.CPUInfo = &task.CPUInfo{
-		Count:      int32(vcpus),
-		CPU:        cpuinfo[0].CPU,
-		VendorID:   cpuinfo[0].VendorID,
-		Family:     cpuinfo[0].Family,
-		PhysicalID: cpuinfo[0].PhysicalID,
+	cpuinfo, err := cpu.Info()
+	vcpus, err := cpu.Counts(true)
+	if err == nil {
```
nravic, line 182: I think you can avoid this + a lot of err == nil and err checking in the function if you declare `var state task.ProcessState` at the top of the func, then instead of returning nil it'll return an empty struct? I don't remember why this was set to be a pointer in the first place

nravic (approved):
lgtm, quick q

pkg/api/state.go
yashanand1910: Great idea.

yashanand1910: Oh, turns out we need to return pointe because that struct contains a lock. It's a proto object.

PR #326, opened 2024-09-05
fix: stop cedana before installing cedana

bsmithai:
## Describe your changes
- Add a stop cedana service portion before copying cedana binary

The purpose of this is to be able to upgrade cedana on a node that already has cedana just by stopping the service temporarily and copying over a new binary before starting a new systemctl service

### DIFF
--- a/cmd/scripts/k8s/setup-host.sh
+++ b/cmd/scripts/k8s/setup-host.sh
@@ -1,6 +1,17 @@
 
 #!/bin/bash
 
+chroot /host /bin/bash <<'EOT'
+APP_NAME="cedana"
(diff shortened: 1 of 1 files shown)


cmd/scripts/k8s/setup-host.sh
```diff
@@ -1,6 +1,17 @@
 
 #!/bin/bash
 
+chroot /host /bin/bash <<'EOT'
```
nravic, line 4: is this 'EOT' right?

nravic (changes requested):
quick question

cmd/scripts/k8s/setup-host.sh
bsmithai: yes

PR #334, opened 2024-09-11
feat: add queue cmd & concurrent tests

swarnimarun:
## Describe your changes

Concurrent tests, need to test correct configuration and fix any bugs or issues.
Also currently podname is broken fix that. One issue I have faced is that internally we use container name to fetch the id from runc, but we can't do that with the regression tests. As there are no pods, and we need to be able to provide id instead of container name.
Still designing this part out a bit.  

## Issue ticket number

Fixes CED-657


swarnimarun:
need to redo it, closing will open a new one

PR #338, opened 2024-09-12
fix: noisy stdout asr output

swarnimarun:
### motivation
very noisy stdout of pod after asr on k8s.

### goal
- makes asr less frequent now down to 60 seconds from 1 second.
- makes GetContainerInfo method trace log rather than debug log. As by default log level is debug, we don't get trace logs in stdout. use "CEDANA_LOG_LEVEL=trace" to adjust this.

### DIFF
--- a/pkg/api/server.go
+++ b/pkg/api/server.go
@@ -377,7 +377,12 @@ func loggingUnaryInterceptor(serveOpts ServeOpts, machineID string) grpc.UnarySe
 			attribute.Bool("server.opts.gpuenabled", serveOpts.GPUEnabled),
 		)
 
-		log.Debug().Str("method", info.FullMethod).Interface("request", req).Msg("gRPC request received")
+		// log the GetContainerInfo method to trace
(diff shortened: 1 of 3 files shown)


pkg/api/server.go
```diff
@@ -377,7 +377,12 @@ func loggingUnaryInterceptor(serveOpts ServeOpts, machineID string) grpc.UnarySe
 			attribute.Bool("server.opts.gpuenabled", serveOpts.GPUEnabled),
 		)
 
-		log.Debug().Str("method", info.FullMethod).Interface("request", req).Msg("gRPC request received")
+		// log the GetContainerInfo method to trace
+		if !strings.Contains(info.FullMethod, "TaskService/GetContainerInfo") {
```
swarnimarun, line 381:
```suggestion
		if strings.Contains(info.FullMethod, "TaskService/GetContainerInfo") {
```

nravic (approved):
lgtm

PR #341, opened 2024-09-13
Cleanup: CLI flags and config

yashanand1910:
## Describe your changes

Flag updates:
- The `--kata` flag for starting daemon is now called `--vsock` because all it does is start the daemon with vsock enabled.
- Add `--port` flag to specify port the daemon should listen on. This will be needed for customer integrations, like Beam. Default: 8080. Also uses this same port when started in vsock mode.
- Add `--port` grandparent flag for all  other CLI commands, for connecting to the daemon.
- Cadvisor will only start if the `--metrics-enabled` flag is passed to the daemon.
- Job service will only start if the `--job-service` flag is passed to the daemon.

Script updates:
- Adds the `--k8s` flag to `build-start-daemon.sh`. that enables metrics and job service by default in the daemon.

Other minor stuff:
- Otel won't even be initialized, if `CEDANA_OTEL_ENABLED` is unset.
- Removes `cmd/asr.go`.
- Removes all traces of `CEDANA_IS_K8S` env var from daemon. Just use `--k8s` flag in the build-start-daemon script, to enable daemon features meant for k8s.

## Issue ticket number
CED-681

Potentially, if these flags are being assumed to be true anywhere. They are false by default now.


cmd/asr.go
nravic, line 1: delete file

nravic (approved):
looks great!

PR #344, opened 2024-09-16
feat: asr support for crio

swarnimarun:
### Describe your changes

Adds crio support. It's doesn't support diskio currently as I am also trying to figure out if or how we can convert it into a percent like metric.

### Issue ticket number

Fix CED-660


nravic (approved):
lgtm!

PR #347, opened 2024-09-18
Log on GPU controller startup poll error

yashanand1910:
### Current behavior/issue

Currently, even if the startup poll RPC call fails, we continue to wait for timeout without printing the status of the RPC poll.

### Proposed solution

Should still be same, but with debug msg including any error msgs.

## Issue ticket number
CED-692


nravic (approved):
lgtm

PR #356, opened 2024-09-25
feat: revert cadvisor simplifications

swarnimarun:
### Describe your changes

This pr tests cadvisor with all it's features enabled, with just the bare minimum API changes.


swarnimarun:
not needed anymore.

PR #357, opened 2024-09-26
fix: remove daemon asr reporting when metrics disabled

swarnimarun:
We are getting failure logs for this in benchmarks and tests. This will ensure we don't anymore.


nravic (approved):
Nice, and then w/ the k8s-helper stuff since we pass --metrics in it's always enabled in k8s, right?

yashanand1910:
> Nice, and then w/ the k8s-helper stuff since we pass --metrics in it's always enabled in k8s, right?

Correct.

PR #358, opened 2024-09-26
fix: add nftables dep

swarnimarun:
nftables dep is broken for apt, this breaks new installs.

### DIFF
--- a/cmd/scripts/k8s/setup-host.sh
+++ b/cmd/scripts/k8s/setup-host.sh
@@ -22,26 +22,24 @@ env \
     CEDANA_OTEL_ENABLED="$CEDANA_OTEL_ENABLED" \
     CEDANA_OTEL_PORT="$CEDANA_OTEL_PORT" \
     CEDANA_LOG_LEVEL="$CEDANA_LOG_LEVEL" \
+    SKIPSETUP="$CEDANA_SKIPSETUP" \
     chroot /host /bin/bash <<'EOT'
 
 if [[ $SKIPSETUP -eq 1 ]]; then
     cd /
-    CEDANA_LOG_LEVEL="$CEDANA_LOG_LEVEL" CEDANA_URL="$CEDANA_URL" CEDANA_AUTH_TOKEN="$CEDANA_AUTH_TOKEN" ./build-start-daemon.sh --systemctl --no-build --otel --k8s
+    ./build-start-daemon.sh --systemctl --no-build --otel --k8s
     exit 0
 fi
 
 # Define packages for YUM and APT
 YUM_PACKAGES=(
-    wget git gcc make libnet-devel protobuf protobuf-c protobuf-c-devel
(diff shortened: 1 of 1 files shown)


cmd/scripts/k8s/setup-host.sh
```diff
@@ -22,26 +22,24 @@ env \
     CEDANA_OTEL_ENABLED="$CEDANA_OTEL_ENABLED" \
     CEDANA_OTEL_PORT="$CEDANA_OTEL_PORT" \
     CEDANA_LOG_LEVEL="$CEDANA_LOG_LEVEL" \
+    SKIPSETUP="$CEDANA_SKIPSETUP" \
     chroot /host /bin/bash <<'EOT'
 
 if [[ $SKIPSETUP -eq 1 ]]; then
     cd /
-    CEDANA_LOG_LEVEL="$CEDANA_LOG_LEVEL" CEDANA_URL="$CEDANA_URL" CEDANA_AUTH_TOKEN="$CEDANA_AUTH_TOKEN" ./build-start-daemon.sh --systemctl --no-build --otel --k8s
```
nravic, line 29: why do we ignore these variables if we're skipping setup?

nravic (commented):
comment

cmd/scripts/k8s/setup-host.sh
swarnimarun: we don't use them for normal startup

swarnimarun: they should already be present

cmd/scripts/k8s/setup-host.sh
```diff
@@ -22,26 +22,24 @@ env \
...
 fi
 
 # Define packages for YUM and APT
 YUM_PACKAGES=(
-    wget git gcc make libnet-devel protobuf protobuf-c protobuf-c-devel
-    protobuf-c-compiler protobuf-compiler protobuf-devel python3-protobuf
-    libnl3-devel libcap-devel libseccomp-devel gpgme-devel btrfs-progs-devel
-    buildah criu
+    wget git gcc make libnet-devel protobuf protobuf-c protobuf-c-devel protobuf-c-compiler protobuf-compiler protobuf-devel python3-protobuf libnl3-devel
+    libcap-devel libseccomp-devel gpgme-devel btrfs-progs-devel buildah criu libnftables1
```
yashanand1910, line 37: Can instead just pull CRIU from https://download.opensuse.org/repositories/devel:/tools:/criu/xUbuntu_22.04/amd64/criu_3.19-4_amd64.deb? As we do in our docker images. @nravic Should be fine if this has been working Ok. I thought there was issues w/ the apt build.

yashanand1910 (approved):

cmd/scripts/k8s/setup-host.sh
swarnimarun: SKIPSETUP was never set to 1 for default installs

PR #359, opened 2024-09-26
fix: add back arm64 support

swarnimarun:
Forgot to add back arm support after the previous revert due to setup bug elsewhere.


nravic (approved):
lgtm

yashanand1910 (approved):

PR #361, opened 2024-09-27
fix: use healthcheck instead of dial

swarnimarun:
Dial API may get removed in a future Go version. Consider switching to using healthcheck instead.


nravic (approved):
lgtm

PR #362, opened 2024-09-30
feat: add setup host scripts test ci

swarnimarun:
### Motivation

We currently don't test setup host scripts.

### Tasks

- Add CI for testing setup host script.
- Split the CI script out of string, provides linting and auto-complete.


bsmithai (approved):
lgtm after all the spell check stuff is resolved

PR #364, opened 2024-10-02
feat: add CEDANA_HELPER_PORT

rahulk789:
## Describe your changes
Add CEDANA_HELPER_PORT env in scripts


build-start-daemon.sh
```diff
@@ -104,6 +105,7 @@ Environment=CEDANA_GPU_CONTROLLER_PATH=$CEDANA_GPU_CONTROLLER_PATH
 Environment=CEDANA_PROFILING_ENABLED=$CEDANA_PROFILING_ENABLED
 Environment=CEDANA_OTEL_ENABLED=$CEDANA_OTEL_ENABLED
 Environment=CEDANA_OTEL_PORT=$CEDANA_OTEL_PORT
+Environment=CEDANA_PORT=$CEDANA_PORT
```
swarnimarun, line 108: we don't pick CEDANA_PORT from env use `--port` instead

swarnimarun (changes requested):
requires --port for start command.

build-start-daemon.sh
rahulk789: At the time of making this, I thought we needed this for the controller to pick up the env. But I just realised that the secrets in helm-chart must suffice. If that's right we can close this pr right? because I've added --port in helm-chart templates.

swarnimarun: --port is required when starting the daemon.

```
cedana daemon start --port ${CEDANA_PORT:8080}
```

swarnimarun: this need to add at the bottom of build-start-daemon script.

rahulk789: I hope this is fixed after the last commit.

yashanand1910:
@swarnimarun @rahulk789 Is this good to merge?

PR #367, opened 2024-10-03
Bump buildah version and fix rootfs diff base paths

bsmithai:
## Describe your changes
- Bump buildah version to latest stable
- Fixes all concurrency issues until we move off buildah cli
- Fix rootfs diff base paths when untar-ing to working container

## Issue ticket number
[CED-711](https://linear.app/cedana/issue/CED-711/bug-concurrent-crio-rootfs-checkpoint-leads-to-using-the-same-buildah)

Looking into adding a test for rootfs diff that makes sure the changed files are where they are supposed to be


pkg/api/crio/crio.go
```diff
@@ -419,6 +421,16 @@ func removeAllContainers() {
 	log.Debug().Msgf("Finished removing all Buildah containers.")
 }
 
+func removeContainer(containerID string) error {
```
nravic, line 424: this should use /cedana/bin no?

nravic (changes requested):

pkg/api/crio/crio.go
bsmithai: Yeah it's the first in the path for systemd service PATH variable

nravic (approved):
lgtm

PR #370, opened 2024-10-08
refactor: asr metrics

swarnimarun:
Removes non-working stuff and tries to add metrics that are more representative.

DiskIO is still zero, I am not sure if it's even meaningful now that FsIoTime exists.

MaxMemory and LoadAvg have been removed.

Accelerator memory might not work on pods we test on (requires nvml on the host). But it's there in case nvml is present and we can get some monitoring information from it before we start shipping gpu-controller to clusters.

Processes is a string as I want to see if that works with BigQuery and our setup, I am considering using a library and for string integer types to avoid precision loss. Could you take a look @nravic, in case it doesn't work well we can move to uint64 instead.

Fixes CED-722


pkg/api/asr.go
```diff
@@ -144,26 +144,45 @@ func (s *service) GetContainerInfo(ctx context.Context, _ *task.ContainerInfoReq
...
-				NetworkIO:     float64(c.Network.RxBytes+c.Network.TxBytes) / 1.,
-				DiskIO:        0,
-				Image:         container.Spec.Image,
-				Labels:        labels,
+				CpuTime:           float64(c.Cpu.Usage.Total) / 1000000000.,
+				FilesystemIoTime:  cumulativeFsTime(c.Filesystem),
+				AcceleratorMemory: cumulativeAcceleratorsMem(c.Accelerators),
+				CurrentMemory:     float64(c.Memory.Usage) / (1024. * 1024.),
+				NetworkIO:         float64(c.Network.RxBytes + c.Network.TxBytes),
+				DiskIO:            cumulativeDiskIoTime(c.DiskIo.IoTime),
```
swarnimarun, line 152:
```suggestion
```

pkg/api/asr.go
```diff
@@ -144,26 +144,45 @@ func (s *service) GetContainerInfo(ctx context.Context, _ *task.ContainerInfoReq
...
+		// time in seconds
+		sum += float64(s.IoTime) / 1000
+	}
+	return sum
+}
+
+func cumulativeDiskIoTime(stats []v1.PerDiskStats) float64 {
+	return 0
+}
+
```
swarnimarun, line 185:
```suggestion

```

nravic (approved):
lgtm

PR #374, opened 2024-10-13
feat: push docker image to cedana-helper-test

rahulk789:
## Describe your changes
Improve dev speed by automating docker builds for cedana and push into cedana/cedana-helper-test repo. Developer can then simply replace the old image in k8 with new built image with respective tags. This will help in debugging errors in  pr.


.github/workflows/pr.yml
```diff
@@ -415,3 +415,44 @@ jobs:
         with:
           filePath: /src/summary.md
           comment_tag: execution
+
+  docker-image-push:
+    name: Docker Image push to cedana-helper-test
+    runs-on: ubicloud-standard-8
```
swarnimarun, line 421: hmm.. do we need 8 or would 2 be enough, I don't think build performance will be affected much, could you test it.

```suggestion
    runs-on: ubicloud-standard-2
```

PR #377, opened 2024-10-18
CED-745: moved some stuff around, added mem for pricing

nravic:
## Describe your changes
Added check for CEDANA_URL, made sure endpoint for gcloud service account was valid, added some more attributes to pubsub data. 

### DIFF
--- a/cmd/daemon.go
+++ b/cmd/daemon.go
@@ -17,6 +17,7 @@ import (
 	"github.com/cedana/cedana/pkg/api/services/task"
 	"github.com/cedana/cedana/pkg/utils"
 	"github.com/rs/zerolog/log"
+	"github.com/shirou/gopsutil/v3/mem"
 	"github.com/spf13/cobra"
 	"github.com/spf13/viper"
 )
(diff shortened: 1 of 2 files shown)


cmd/daemon.go
```diff
@@ -110,7 +112,7 @@ func gcloudAdcSetup(ctx context.Context) error {
 		// already present skip
 		return nil
 	}
-	url := getenv("CEDANA_SERVER_URL", "https://sandbox.cedana.ai") + "/k8s/gcloud/serviceaccount"
+	url := getenv("CEDANA_URL", "https://sandbox.cedana.ai") + "/k8s/gcloud/serviceaccount"
```
yashanand1910, line 115: Should use viper to fetch this value instead of directly from env. It might exist in the config.

cmd/daemon.go
```diff
@@ -165,6 +173,11 @@ func pollForAsrMetricsReporting(ctx context.Context, port uint32) {
 			// Publish a message
 			result := topic.Publish(ctx, &pubsub.Message{
 				Data: b,
+				Attributes: map[string]string{
```
yashanand1910, line 176: Is this data not fetchable from Cadvisor itself?

yashanand1910 (changes requested):
Small fixes needed. Otherwise lgtm.

bsmithai (approved):
lgtm

PR #380, opened 2024-10-24
refactor: gpu binary downloads

swarnimarun:
### Describe your changes
Downloads were broken if you added the path, now it will ensure the download happens even if the path is specified. Also does them in parallel for perf & adds gpu flag if on nvidia supported node.

Ideally we will do all the setup in setup host but this improves the user experience for the end user. Also makes the logs for download "info" so it will be outputted to the user stdout by default.

Fixes CED-730


yashanand1910 (approved):
Looks good!

nravic (approved):
lgtm

PR #395, opened 2024-11-08
fix: runc query bug

swarnimarun:
### Fix for DevZero.

This is a runc query bug. Tested the PR, the said bug affects all endpoints except specifically `/restore` lmao.

The point is, if it's container not found error then we still append the container to the list of containers we return with runcQuery containers. Doing a continue for that case should fix the issue.

Though we also need to return an error if it's an empty list as that is what use to continue the loop on controller rest api side.

This regression was introduced in:  https://github.com/cedana/cedana/pull/363

This is in-fact quite easy to reproduce.

The method to reproduce it is,
- Use it on a multi node cluster (on a single node cluster you can try making a request with wrong container name, and it will give error message without runcDiscoveryErr). But technically, we can't hit this bug on single node clusters.
- Now we can just make multiple requests with containers names that are on different nodes and one of the requests will hit the wrong node first.
- When that happens we don't return an error from this function now, so the search loop breaks. 

This is also the same semantics I follow in `https://github.com/cedana/controller/pull/54` so that PR will also be broken, technically most APIs will break.

The cause is simply we don't have a well-defined API guidance for now, and any changes to API behaviors will break our users/consumers that includes ourselves.

I am adding a note for making changes for improving APIs on controller side to catch these issues and snapshot testing a priority with those these issues should be much easier to catch. Whenever our input or output behaviours change.

Also, we need to improve error messages in general, as well.

cc: @nravic @bsmithai

### DIFF
--- a/pkg/api/runc.go
+++ b/pkg/api/runc.go
@@ -346,7 +346,6 @@ func (s *service) RuncRestore(ctx context.Context, args *task.RuncRestoreArgs) (
 func (s *service) RuncQuery(ctx context.Context, args *task.RuncQueryArgs) (*task.RuncQueryResp, error) {
 	var containers []*task.RuncContainer
 	if len(args.ContainerNames) == 0 {
-
 		runcContainers, err := runc.RuncGetAll(args.Root, args.Namespace)
 		if err != nil {
 			return nil, status.Error(codes.NotFound, err.Error())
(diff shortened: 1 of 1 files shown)


pkg/api/runc.go
```diff
@@ -381,6 +382,10 @@ func (s *service) RuncQuery(ctx context.Context, args *task.RuncQueryArgs) (*tas
 		})
 	}
 
+	if len(containers) == 0 {
+		return nil, status.Error(codes.NotFound, "no containers found")
+	}
+
```
swarnimarun, line 388:
```suggestion
```

swarnimarun: Note this change was required as in crioRootfsCheckpoint we do,

```go
			ctrResp, runcDiscoveryErr := cts.TaskService.RuncQuery(ctx, ctrByNameArgs)
			if runcDiscoveryErr == nil {
				runcId = ctrResp.Containers[0].ID
				break
			} else {
				logger.Error().Err(runcDiscoveryErr).Msg("error querying runc")
				fmt.Printf("Error: %s", runcDiscoveryErr)
			}
```

Note this is the actual fix here, as we unconditionally set runcId if there is not an error.

nravic (approved):
lgtm

PR #396, opened 2024-11-13
Implement remote DB

yashanand1910:
## Describe your changes
Also see https://github.com/cedana/cedana-api/pull/37.

Currently, the managed jobs information is stored in a local DB. CEDANA_REMOTE usage is outdated.

Proposed solution
- Make it so that when CEDANA_REMOTE=true, the remote DB is used through the propagator.
- Remove all existing instances of CEDANA_REMOTE usage, including code that uploads/downloads checkpoints as the daemon should not have this complexity. This will all eventually be done through the propagator/CDN.
- The DB will be managed by the propagator.

## Issue ticket number
CED-757

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -12,19 +12,9 @@ on:
         description: "Run build with debugging enabled"
         required: false
         default: false
-      debug_smoke_test:
-        type: boolean
-        description: "Run smoke test with debugging enabled"
-        required: false
-        default: false
       debug_regression_test:
         type: boolean
         description: "Run regression test with debugging enabled"
-        required: false
-        default: false
-      debug_correctness_test:
-        type: boolean
-        description: "Run correctness test with debugging enabled"
         required: false
         default: false
       debug_unit_test:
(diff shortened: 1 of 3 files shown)


.gitmodules
```diff
@@ -1,4 +1,4 @@
 [submodule "pkg/api/services/cedana-api"]
 	path = pkg/api/services/cedana-api
 	url = https://github.com/cedana/cedana-api
-  branch = main 
+  branch = feature/ced-757-cedana_remote-to-use-remote-job-db
```
nravic, line 4: reminder to swap out

pkg/api/server.go
```diff
@@ -108,13 +105,19 @@ func NewServer(ctx context.Context, opts *ServeOpts) (*Server, error) {
 		}
 	}
 
+	var database db.DB
+	if viper.GetBool("remote") {
+		database = db.NewRemoteDB(ctx, viper.GetString("connection.cedana_url")+"/k8s/jobs")
```
nravic, line 110: makes sense to me

nravic (commented):
Looks good - things are failing in tests for some reason. I'm going to add the propagator side. Will approve after tests pass, I add propagator side and maybe we can add a remoting c/r test to regression.

PR #411, opened 2024-11-21
refactor: restart management & cleanup

swarnimarun:
### Description

Restarts from helper are currently broken as they don't use the initial setup commands.
Also cleanup didn't manage things correctly.

Further we now avoid using the root `/` path, instead we use the `/cedana/scripts` directory for storing scripts and executing them. This ensure we don't overwrite or conflict with any existing files.

Fixes CED-782


cmd/scripts/k8s/cleanup-host.sh
```diff
@@ -1,23 +1,4 @@
...
-
-systemctl stop cedana.service
-
-
-rm -f /usr/local/bin/cedana
-rm -f /build-start-daemon.sh
-EOT
-
-echo "Clean up completed."
+chroot /host /bin/bash /reset.sh
```
swarnimarun, line 3:
```suggestion
chroot /host /bin/bash /cedana/scripts/reset.sh
```

.github/workflows/pr.yml
```diff
@@ -136,8 +136,10 @@ jobs:
         run: |
           sudo chmod +x ./cedana
           sudo cp ./cedana /usr/local/bin/
-          sudo cp ./build-start-daemon.sh /build-start-daemon.sh
-          sudo ./setup-host.sh
+          mkdir -p /cedana/scripts
```
swarnimarun, line 139:
```suggestion
          sudo mkdir -p /cedana/scripts
```

pkg/container/utils_linux.go
```diff
@@ -256,7 +256,7 @@ func Create(root, id string, config *configs.Config) (*RuncContainer, error) {
 	if cm.Exists() {
 		// pids, err := cm.GetAllPids()
 		// Reading PIDs can race with cgroups removal, so ignore ENOENT and ENODEV.
-		if err != nil && !errors.Is(err, os.ErrNotExist) && !errors.Is(err, unix.ENODEV) {
+		if !errors.Is(err, os.ErrNotExist) && !errors.Is(err, unix.ENODEV) {
```
yashanand1910, line 259: This will always return err.

yashanand1910 (changes requested):
Check comments - cause of regression failure.

pkg/container/utils_linux.go
swarnimarun: I think this was a golint suggestion... reverting let's see.

swarnimarun, line 259:
```suggestion
		if err != nil && !errors.Is(err, os.ErrNotExist) && !errors.Is(err, unix.ENODEV) {
```

cmd/scripts/k8s/setup-host.sh
```diff
@@ -4,20 +4,27 @@ chroot /host /bin/bash <<'EOT'
...
+    echo -n > /var/log/cedana-daemon.log
+else
+    echo "No cedana-daemon.log file found"
+    echo "Skipping $APP_NAME service stop"
+fi
 EOT
 
+# Create Cedana directories
+mkdir -p /host/cedana /host/cedana/bin /host/cedana/scripts
+
```
swarnimarun, line 21:
```suggestion

# We load the binary from docker image for the container
```

nravic (approved):
lgtm

PR #421, opened 2024-11-27
feat: handle automatic shim setup

swarnimarun:
Shim setup should work, currently testing this PR properly.

Also we download criu directly from cloudsmith we should move that to propagator so that we can control which version we ship as latest or stable similar to shim.

fixes CED-782

edit: do not merge (on 27th Nov 2024), ref https://github.com/cedana/cedana/pull/421#issuecomment-2504663989, need to merge my changes in helm chart, ideally it won't break anything but still might cause issues.

only testing on aws eks and gke


setup-host.sh
```diff
@@ -74,7 +74,51 @@ fi
 
 GPU=""
 if [ -f /proc/driver/nvidia/gpus/ ]; then
-    echo "nvidia-gpu found!"
+    echo "Detected NVIDIA GPU! Ensuring CUDA drivers are installed..."
+    if [ /sbin/ldconfig -p | grep -q libcuda.so.1 ]; then
```
swarnimarun, line 78:
```suggestion
    if $(/sbin/ldconfig -p | grep -q libcuda.so.1); then
```

swarnimarun:
I found a bug in it will fix, but more importantly we need to make a minor change to the helm chart before this works entirely as expected. And we can use the "cedana" runtime class in kubernetes readily.

swarnimarun:
closing in favour of #426

PR #422, opened 2024-12-02
Deal with changed order of restore for cedana-image-streamer

lianakoleva:


pkg/api/restore.go
```diff
@@ -45,8 +45,8 @@ const (
 	KATA_TAR_FILE_RECEIVER_PORT  = 9998
 )
 
-func (s *service) setupStreamerServe(ctx context.Context, dumpdir string, num_pipes int32) *exec.Cmd {
-	cmd := exec.CommandContext(ctx, "cedana-image-streamer", "--dir", dumpdir, "--num-pipes", fmt.Sprint(num_pipes), "serve")
+func (s *service) setupStreamerServe(dumpdir string, num_pipes int32) {
+	cmd := exec.Command("cedana-image-streamer", "--dir", dumpdir, "--num-pipes", fmt.Sprint(num_pipes), "serve")
```
nravic, line 49: why remove context here?

nravic (commented):
added a comment, lgtm otherwise!

pkg/api/restore.go
lianakoleva: i understand that it helps with streamer to exit if daemon exits, but don't understand why this isn't done elsewhere.

nravic: Yep - I think we should be using CommandContext everywhere applicable, so that things are cleaned up nicely and we can start enforcing timeouts. Of those 17 instances, the ones that are in the hot path and can take time (i.e not be one-off commands that complete quickly) should all be changed to CommandContext.

lianakoleva: ok i will revert this one

nravic (approved):
lgtm!

PR #426, opened 2024-12-04
feat: handle automatic shim dl

swarnimarun:
Replaces PR #421 

fixes CED-782

### DIFF
--- a/cmd/helper.go
+++ b/cmd/helper.go
@@ -73,6 +73,7 @@ var helperCmd = &cobra.Command{
 
 		port, _ := cmd.Flags().GetUint32(portFlag)
 		startChroot, _ := cmd.Flags().GetBool("start-chroot")
+    startChroot = startChroot || setupHost
 		startHelper(ctx, startChroot, port)
 
 		return nil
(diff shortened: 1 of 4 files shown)


cmd/helper.go
```diff
@@ -73,6 +73,7 @@ var helperCmd = &cobra.Command{
 
 		port, _ := cmd.Flags().GetUint32(portFlag)
 		startChroot, _ := cmd.Flags().GetBool("start-chroot")
+    startChroot = startChroot || setupHost
```
swarnimarun, line 76:
```suggestion
        startChroot = startChroot || setupHost
```

cmd/helper.go
```diff
@@ -73,6 +73,7 @@ var helperCmd = &cobra.Command{
 
 		port, _ := cmd.Flags().GetUint32(portFlag)
 		startChroot, _ := cmd.Flags().GetBool("start-chroot")
+        startChroot = startChroot || setupHost
```
nravic, line 76: is startChroot used at all anymore? The func can probably just go

nravic (approved):
lgtm, added small question

cmd/helper.go
swarnimarun: It's used only in a helper function to do a single call with chroot. It's never passed as a flag anywhere in k8s workflow.

I will make tickets today to clean up k8s helper part of the project.

PR #431, opened 2024-12-18
Full redesign using Plugins (v1.0)

yashanand1910:
This is a full rewrite of the daemon and CLI, following a highly-modular architecture using [Go plugins](https://pkg.go.dev/plugin), and the well-known adapter pattern used in HTTP libraries. It implements 100% of the functionality using just 50% lines of code. The goals of this refactor are to:
- Maximize code re-use for functionally identical parts of the design, for e.g. compressing dump files.
- Rid daemon of use-case specific logic (k8s or runtime-specific), and make it possible to plug it in if needed. The daemon alone should only contain the bare-minimum process C/R and connectivity to Cedana's systems.
- Reduce the size of binary, and dependencies required to run bare-minimum C/R.
- Make it easier to add support for newer runtimes. Ideally, it should be easy for the developers of a runtime to add support for their runtime themselves. Currently, this is very hard to do as the code is hard to follow.
- Make it possible to add customer-specific logic without affecting the daemon.
- Make it easier to add other managed C/R support, for e.g. for VMs.
- Unlock potential for serious optimizations, without worrying about breakage.
- Allow more focused async work while minimizing merge conflicts.

**When reviewing, I recommend reading the code in an IDE instead of looking at the diff, as the diff is quite large. Please go through the below items before you start, to get an idea of what's changed.**

## What's changed
- [Simplified build system](#simplified-build-system)
- [Reorganized source tree](#reorganized-source-tree)
- [Plugin system](#plugin-system)
  - [Build & install](#build--install)
  - [Design](#design)
  - [Features](#features)
- [New DB interface](#new-db-interface)
- [Advanced health check system](#advanced-health-check-system)
- [Improved profiling](#improved-profiling)
- [Upgraded tests including parallelism](#upgraded-tests-including-parallelism)
- [Simplified API](#simplified-api)
- [Refreshed config](#improved-config)
- [Accompanying changes](#accompanying-changes)
- [What's left](#whats-left)

## Simplified build system
Our building needs are simple enough to be handled by a single Makefile. Scripts like `build.sh`, `build-start-daemon.sh`, `reset.sh` have been removed and are now part of the Makefile. Try `make help` to see the available options:
```
Usage:
  all                       Build and install (with all plugins)

Cedana
  build                     Build the binary
  install                   Install the binary
  start                     Start the daemon
  stop                      Stop the daemon
  start-systemd             Build and start the systemd daemon
  stop-systemd              Stop the systemd daemon
  reset                     Reset (everything)
  reset-db                  Reset the local database
  reset-config              Reset configuration files
  reset-tmp                 Reset temporary files
  reset-logs                Reset logs

Plugins
  plugins                   Build plugins
  plugins-install           Install plugins
  reset-plugins             Reset & uninstall plugins

Testing
  test                      Run all tests
  test-unit                 Run unit tests (with benchmarks)
  test-regression           Run all regression tests (PARALLELISM=<n>)
  test-regression-cedana    Run regression tests for cedana
  test-regression-plugin    Run regression tests for a plugin (PLUGIN=<plugin>)

Docker
  docker-test               Build the test Docker image
  docker-test-push          Push the test Docker image

Utility
  format                    Format all code
  help                      Display this help
```

## Reorganized source tree
The source tree has been reorganized to follow [Go standards](https://github.com/golang-standards/project-layout). Particularly, the `internal` directory contains all packages not meant to be imported by other packages. The `pkg` directory contains packages that can be imported by other packages. The `cmd` directory contains CLI code. Same is followed for each plugin, which also exist in the same tree:
```
.
├── cmd
│   ├── attach.go
│   ├── daemon.go
│   ├── dump.go
│   ├── restore.go
│   └── ...
├── internal
│   ├── db
│   ├── metrics
│   └── server
├── pkg
│   ├── config
│   ├── criu
│   ├── features
│   ├── flags
│   ├── io
│   └── ...
├── plugins
│   ├── containerd
│   │   ├── cmd
│   │   ├── internal
│   │   ├── pkg
│   ├── crio
│   │   ├── cmd
│   │   ├── internal
│   │   ├── pkg
│   └── runc
│       ├── cmd
│       ├── internal
│       ├── pkg
└── ...
```
A plugin can also be on a separate repository or in its own Go module. However, to avoid compatibility issues involving dependencies (especially with generated proto code), we let Go handle everything in a single module.

### Build & install
Each plugin is built as a separate binary with the name `libcedana-<plugin>.so` (use `make plugins`). The main binary automatically loads any plugins found in `/usr/local/lib`. 

The new subcommand `cedana plugin ...` can be used to manage plugins:
```sh
Usage:
  cedana plugin [command]

Available Commands:
  features    Show feature matrix of plugins
  install     Install a plugin
  list        List plugins
  remove      Remove a plugin
```
Run `cedana plugin list`:

image

Currently, since we don't have an online registry, invoking these commands will search for _locally built_ binaries in paths specified by the env var `CEDANA_PLUGINS_LOCAL_SEARCH_PATH` (e.g. `/path/to/dir1:/path/to/dir2`). The Makefile already sets this to the root of the source tree, where all binaries are built, so you can simply use `make plugins-install` to install all the locally built plugins.

### Design
The design mostly follows what's illustrated below. Below is a simplified runtime view of invoking `cedana dump runc ...`:

image

1. The subcommand `cedana dump runc ...` is only available if the runc plugin is exporting the `DumpCmd` symbol (check `plugins/runc/main.go`). The runc plugin only sets the _specific_ flags it needs (such as `--id`, `--root`), while the parent cmd handles all the common flags, and sending the request to the daemon.
2. The daemon receives the request, and runs it through a list of adapters, before finally sending it to CRIU. If the request's `type` is `runc`, it will use the `DumpMiddleware` exported by the runc plugin and plug it in the adapter chain. See `internal/server/dump.go`: https://github.com/cedana/cedana/blob/24b37c64d3f51630fc193640d0df92080fed0d4a/internal/server/dump.go#L23-L44
3. This way, the runc plugin only implements the specifics of the runc runtime, while the daemon handles the common logic, and invoking CRIU.
4. The same pattern is followed for `dump`, `restore`, `run`, and `manage`.

### Features
Symbols that can be exported by a plugin are well-typed and are defined in `pkg/features/supported.go`. A `feature` implements a convenient method called `IfAvailable(do func(), filter ...string)`, which is the _only_ method you will ever need to access a feature exported by a plugin. For a usage example, see: https://github.com/cedana/cedana/blob/3e97d8f5241b53aa40307635771724f0ac8b4b54/internal/server/dump.go#L86-L95

A useful helper command is `cedana features` (alias of `cedana plugin features`), which lists all the features supported by all the plugins. This is useful for debugging, when implementing a new feature, or when you want to know what a plugin supports. Use the `--errors` flag to also output all errors encountered while loading the plugins.

image

## New DB interface
The DB interface has new methods: https://github.com/cedana/cedana/blob/24b37c64d3f51630fc193640d0df92080fed0d4a/internal/db/db.go#L13-L33

Naturally, though not necessarily, this is implemented using separate tables for Jobs, Hosts, and Checkpoints. See `internal/db/sql/schema.sql` for the new local DB schema. 

This allows separation of jobs from hosts and checkpoints. Each checkpoint belongs to a job, and each job belongs to a host.

Since daemons connected to the same `CEDANA_URL` will share the DB (through the propagator), it becomes necessary to specify what host a job belongs to. Running `cedana ps` (alias of `cedana job list`) on a host will only show jobs that belong to that host:

image

Running `cedana ps --all` will show all jobs, regardless of the host, with the `Status` saying `remote` for a remote job:

image

This allows directly restoring a job from a remote host, if a checkpoint is available.

The checkpoints table now becomes are large index of checkpoints, instead of a separate list for each job. To view checkpoints for a job, use `cedana checkpoints <JID>` (alias of `cedana job checkpoint list <JID>`):

image

## Advanced health check system
The health check system has been upgraded to be more flexible and include warnings and more actionable information. The `--full` flag in `cedana daemon check` also health checks all installed plugins. An example output:

image

As explained in [features]($features), each plugin can implement the `HealthChecks` feature to export its own health checks. Look at `pkg/features/supported.go` and `pkg/types/check.go` for details.

## Improved profiling
The profiling system has been upgraded to be more contextual and _invisible_. Profiling data is no longer part of `cedana-api`, but, when enabled, is instead sent as [gRPC Metadata](https://grpc.io/docs/guides/metadata/). To enable, set `Profiling.Enabled=true` in config or `CEDANA_PROFILING_ENABLED=true` in env.

Since each adapter to a request (see [plugin system](#plugin-system)) is a well-defined single-responsibility function, it makes sense to profile each of them. When enabled, the daemon will profile each adapter, and send back flattened data as a gRPC trailer. For readability's sake, this profiling is completely invisible and handled by the adapter logic in `pkg/types/adapter.go` (also see `pkg/types/timer.go`). This flattened data is parsed by the `cmd` package and displayed as shown below. An example out of `cedana dump containerd ...` when profiling is enabled:

image

Above, you can see complete flow of the request before it reaches CRIU's dump function, including _which_ plugin (2nd column) the adapter belongs to. For `containerd`, you can see that the request is largely handled by the low-level runtime `runc`'s plugin. The second table above shows a compressed view of the same data, with only the total time spent in each plugin/category. Note that, above there are some components that are executed concurrently, e.g. `rootfs`, so the time you see in the (flattened) data above is only the time spent _waiting_ for `rootfs` to finish.

In many cases, we may need to add more context to this data, or add more components to it. Helpers defined in `pkg/profiling/timing.go` can be used. A good example is adding a new component to the `compression` category as seen above: https://github.com/cedana/cedana/blob/3e97d8f5241b53aa40307635771724f0ac8b4b54/internal/server/filesystem/dump_adapters.go#L90-L92

These helpers use the passed `context` to store profiling data. If the `context` already has parent profiling data, the data is added as a component to the parent.

Behind the scenes, if OTel is enabled (config `Metrics.Otel=true`), this data is also captured in OTel spans.

## Upgraded tests including parallelism
Tests are now configured to run inside a Docker container, ridding the need for tedious setup scripts. This allows for testing locally, during development, without affecting the environment. 

The test directory now looks like this:
```
test
├── regression
│   ├── basic.bats
│   ├── cr.bats
│   ├── run.bats
│   ├── manage.bats
│   ├── profiling.bats
│   ├── plugins
│   │   ├── containerd.bats
│   │   ├── crio.bats
│   │   ├── gpu.bats
│   │   └── runc.bats
│   ├── helpers
│   │   └── ...
└── workloads
```
Tests are grouped by functionality, and each plugin has its own test file.

Running `make help` you'll see the following test commands:
```sh
Testing
  test                      Run all tests
  test-unit                 Run unit tests (with benchmarks)            # All go tests & benchmarks
  test-regression           Run all regression tests (PARALLELISM=<n>)  # All tests in test/regression
  test-regression-cedana    Run regression tests for cedana
  test-regression-plugin    Run regression tests for a plugin (PLUGIN=<plugin>)
```

When running any of the test commands locally, the tests are automatically run inside a Docker container (using `cedana/cedana-test:latest`). The CI is also configured to use this Docker image.

Each test command above runs the test file  **two times**, in different modes:
1. Unique daemon & DB instance for each test.
2. Single persistent daemon & DB instance across a test file.

This is to allow catching bugs that may arise due to the daemon's state being persisted across tests.

Each test command is also configured to run parallel-y, configured by the `PARALLELISM` variable passed to the test command. E.g. `make test-regression PARALLELISM=4` will run at most 4 tests in parallel at a time.

For mode 1, parallelism offers no benefits apart from fast execution time, as each test is completely isolated. However, for mode 2, parallelism may shed light on bugs in the daemon when it's handling multiple requests concurrently.

## Simplified API
The above changes, and the need for a modular system using plugins, have resulted in a much simpler API. In short, there's no longer a need to have a separate request/response message for each type of request. For e.g., now there is a single `DumpReq` message, which is used for all types of dump requests, regardless of the runtime. The daemon just uses the `Type` to determine if it needs a plugin to handle the request.

```
message DumpReq {
  string Dir = 1;
  bool Stream = 2;
  string Type = 3;
  criu.criu_opts Criu = 4;
  bool External = 5;

  Details Details = 6;
}
```
Please take a look at the accompanying `cedana-api` [PR](https://github.com/cedana/cedana-api/pull/42) for details.

## Refreshed config
The default configuration is now well-typed, and can be found in: https://github.com/cedana/cedana/blob/3e97d8f5241b53aa40307635771724f0ac8b4b54/pkg/config/config.go#L22-L53

As before, this configuration is overridden during runtime, by order of priority:
1. Flag. E.g. `--port` flag when starting the daemon.
2. Environment variable. E.g. `CEDANA_PORT=8080` when starting the daemon.
3. Config file. E.g. `Port=8080` in `~/.cedana/config.json`.
4. Default value.

Environment variables are prefixed with `CEDANA_`, and are in uppercase. E.g. `CEDANA_PORT=8080`. Each config item has a default corresponding environment variable. And, it's possible to specify an environment variable alias for backward compatibility. For e.g., in `pkg/config/types.go`, the `Config.Connection.URL` has the default environment variable counterpart `CEDANA_CONNECTION_URL`. However, an alias `CEDANA_URL` is also specified using the struct tag as shown here: https://github.com/cedana/cedana/blob/3e97d8f5241b53aa40307635771724f0ac8b4b54/pkg/config/types.go#L41-L42

The `env_aliases` tag can have multiple comma-separated values. This is a custom feature, not part of [Viper](https://github.com/spf13/viper) (yet?).

## Accompanying changes
The following PRs are also part of this rewrite:
- [cedana-api](https://github.com/cedana/cedana-api/pull/42): Changes as explained in [simplified API](#simplified-api).
- [cedana-gpu](https://github.com/cedana/cedana-gpu/pull/101): Due to [simplified API](#simplified-api) and [new health check system](#advanced-health-check-system), the GPU controller has been updated. A small additional change to allow sending a `SIGUSR1` signal to the intercepted process if the GPU controller prematurely exits.
- [cedana-propagator](https://github.com/cedana/cedana-propagator/pull/19): Plugin & binary downloads, and DB updates for remote DB.

### Other minor changes
- Kubernetes specific commands has been moved to its own plugin (see `plugins/k8s`). It's produced as `libcedana-k8s.so`.
- Helper commands like `cedana runc get` were removed (including the RPC `RuncQuery`). There's a subcommand `cedana query` that can be used to query for information. Plugins implement their own query subcommands, and query handlers. Through CLI, they can be accessed as `cedana query <plugin> ...`. Try `cedana query k8s --help`.

## What's left
- [ ] Streaming support
- [x] Online plugin registry (through propagator)
- [x] Runc plugin
- [x] Containerd plugin
- [x] K8s plugin
- [ ] CRIO plugin
- [ ] Kata plugin


cmd/plugin.go
```diff
@@ -0,0 +1,441 @@
...
+	// Subcommand flags
+	pluginListCmd.Flags().
+		BoolP(flags.AllFlag.Full, flags.AllFlag.Short, false, "List all available plugins")
+	pluginRemoveCmd.Flags().
+		BoolP(flags.AllFlag.Full, flags.AllFlag.Short, false, "Remove all installed plugins")
+	pluginFeaturesCmd.Flags().
+		BoolP(flags.ErrorsFlag.Full, flags.ErrorsFlag.Short, false, "Show all errors")
+
+	// Add aliases
+	pluginCmd.AddCommand(utils.AliasOf(pluginListCmd, "ls"))
```
nravic, line 35: do we need both ls and list?

yashanand1910: We certainly do not _need_ both. But it’s a quite conventional in most CLI programs (e.g. `ctr`) I have seen to have two types of aliases:
1. Shorthand names (ls for list)
2. Shortcut in the root command for a grandchild command. 

Plus it’s a single line of code for an alias.

Makefile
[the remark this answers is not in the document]
yashanand1910: Can you remove the old binary and build again?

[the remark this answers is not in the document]
yashanand1910: Just `make` should do. It install's it too.

.github/workflows/pr.yml
```diff
@@ -1,32 +1,19 @@
 name: PR
 
 on:
-  pull_request:
-    types: [opened, synchronize]
-    branches:
-      - main
+  push: # TODO: revert after resolving merge conflicts
```
swarnimarun, line 4:
```suggestion
  push:
```

did we do the revert?

.github/workflows/pr.yml
```diff
@@ -135,332 +140,461 @@ jobs:
...
+  #     - name: Run script
+  #       run: |
+  #         sudo chmod +x ./cedana
+  #         sudo cp ./cedana /usr/local/bin/
+  #         sudo mkdir -p /cedana/scripts
+  #         sudo cp ./setup-host.sh /cedana/scripts/setup-host.sh
+  #         sudo cp ./build-start-daemon.sh /cedana/scripts/build-start-daemon.sh
+  #         sudo /cedana/scripts/setup-host.sh
+  #       env:
+  #         CLOUDSMITH_ENTITLEMENT_TOKEN_CRIU: "${{ secrets.CLOUDSMITH_ENTITLEMENT_TOKEN_CRIU }}"
```
swarnimarun, line 169:
```suggestion
```
we can remove the commented code if not needed.

.github/workflows/pr.yml
```diff
@@ -135,332 +140,461 @@ jobs:
...
+  #           type=raw,value=${{ env.GIT_BRANCH }}-${{ env.GIT_HASH }}
+  #           type=raw,value=${{ env.GIT_BRANCH }}
+
+  #     - name: Build and push to cedana-helper-test
+  #       uses: docker/build-push-action@v6
+  #       with:
+  #         push: true
+  #         context: .
+  #         tags: ${{ steps.meta.outputs.tags }}
+  #         labels: ${{ steps.meta.outputs.labels }}
```
swarnimarun, line 600:
```suggestion
```
if not needed we should remove commented code.

swarnimarun (commented):
could you take a look at the ci?

.github/workflows/pr.yml
yashanand1910: No, sorry, I'll be updating the CI. Since there were significant changes I had commented out for testing.

PR #427, opened 2024-12-04
Add direct remoting support in daemon

lianakoleva:
## Describe your changes
Directly write checkpoint files to and read checkpoint files from AWS S3 bucket without writing them to disk (in-memory streaming). 

Usage:
demo

## Checklist before requesting a review
- [X] Have I performed a self-review?
- [x] Have I added regression or unit tests (where it makes sense) for my changes?
- [x] Have I updated the README if changes have resulted in it being out of date?
- [X] Have I checked to ensure my PR only introduces the changes it's intended to? E.g. No extraneous formatting changes.
- [ ] Have I ensured that there are no performance regressions by checking benchmark results? Not working.
- [ ] Is this a breaking change? If yes, please describe areas affected and request reviews from developer stakeholders. No.

PR #434, opened 2025-01-07
fix: gpu restore for runc

swarnimarun:
### Motivation

Bug with gpu runc restore.

Technically it still works, but we start an extra gpu controller and use more shared memory, not desirable, breaks on systems with not large enough dev shm.

This is also the cause for ghost processes.

### Changes

If jid already exists we know it's a managed process which means we can ignore starting another gpu controller.

### DIFF
--- a/pkg/api/gpu.go
+++ b/pkg/api/gpu.go
@@ -150,9 +150,9 @@ func (s *service) StartGPUController(ctx context.Context, uid, gid int32, groups
 		},
 	}
 
-  if _, err := os.Stat("/run/nvidia/driver"); err == nil {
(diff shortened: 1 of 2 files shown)


swarnimarun:
Closing this, #427 broke the k8s setup, I created a new PR without those commits included.

Will fix the k8s stuff for aws script separately. Also currently setup host script CI doesn't test for missing scripts using the dockerimage, if we forget to copy a script to docker image and then to the host system. It will still not be an error.

Created [CED-899](https://linear.app/cedana/issue/CED-899/bug-ci-doesnt-correctly-test-the-docker-image-that-we-ship-for-k8s) for it.

PR #435, opened 2025-01-07
fix: gpu restore runc

swarnimarun:
### Motivation

Bug with gpu runc restore.

Technically it still works, but we start an extra gpu controller and use more shared memory, not desirable, breaks on systems with not large enough dev shm.

This is also the cause for ghost processes.

### Changes

If jid already exists we know it's a managed process which means we can ignore starting another gpu controller.

Fixes CED-889

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -349,6 +349,7 @@ jobs:
     steps:
       - uses: actions/checkout@v4
         with:
+          ref: ${{ github.event.pull_request.head.sha }}
           fetch-depth: 0
           submodules: "recursive"
 
--- a/cmd/daemon.go
+++ b/cmd/daemon.go
@@ -185,11 +185,10 @@ func pollForAsrMetricsReporting(ctx context.Context, port uint32) {
 				},
 			})
 			// Get the server-assigned message ID
-			id, err := result.Get(ctx)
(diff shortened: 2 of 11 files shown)


yashanand1910 (commented):
I don’t understand how a GPU controller is already started outside of cedana? It’s the restore that must start a controller, so the daemon knows which controller is attached to which process.

swarnimarun:
@yashanand1910 controller is not started outside of cedana it's started by `runc manage`. But restore will always try to start a new gpu controller even if it's already running, gpuRestore function didn't use to check it.

yashanand1910:
I’m still confused. When a dump happens, the controller dies. Is that not happening?

swarnimarun:
@yashanand1910 yeah, but remember when we want to perform a restore on a container, you need to start it beforehand.

swarnimarun:
This means the controller is already running on it. This wouldn't happen once use shim to directly store to a container, but as long as the container starts before the restore request we have to check this.

yashanand1910:
Got it. Thanks for clarifying. I think @bsmithai already has restore through shim working, waiting for his PR. I’m approving this. Once #431 is merged, this will get reverted.

yashanand1910 (approved):

PR #441, opened 2025-01-10
fix: [wip] k8s scripts

swarnimarun:
fixes k8s install


swarnimarun:
I know this will break ci I want to just test it for now so opening a PR

PR #447, opened 2025-01-15
[WIP] fix: changes for runc gpu restore support to plugin arch branch

swarnimarun:
WIP

### Tasks:
- [x] Fix issues with setup host scripts.
- [x] Port over logging changes for improved demo logging
- [ ] Make GPU setup changes where required (cc: @yashanand1910 I noticed the /dev/shm mount in the plugins branch has different options from the ones we used for testing gpu before, is that intended).
- [ ] Fix for external mounts to follow symblinks.
- [ ] Registry download doesn't happen for shim, also shim path doesn't match setup-host-script.
- [ ] Figure out any other required changes, especially around JID usage for runcRestore. I am testing it currently, I believe we need it, but the control flow is different so trying to make sure my changes don't break anything else.
- [ ] For Runc manage we want to allow, managing without the container already being started.

Note: "Runc manage change" will probably be required even after the shim restore pr from brandon as we can't directly mount /dev/shm but have to mount the `/dev/shm/cedana-gpu.xxx` file, as runc/criu can't mount a file before it exists.

If we can figure out how to get away with mounting /dev/shm then this change will not be needed.


yashanand1910:
@swarnimarun 
1. One of the options for the /dev/shm mount in there was removed when I noticed runc complaining about how a filesystem-specific option is not valid. These work.
2. Following symlink is not needed anymore. libcedana-gpu.so is statically compiled now (only except libc). We don't need to mount any other library.
3. The shim binary name (as on cloudsmith) just needs to be added in the binary list for the k8s plugin in `pkg/plugins/registry.go`. This registry is stored locally for now, but it's updated during runtime with information from propagator (if you run `cedana plugin list`).
4. For now, we continue to mount /dev/shm. I have tested it locally and it works for restore. I'm not sure about the issues on the `main` branch. Until we have a better way, we should do this. During restore, the `cedana restore job` or `cedana restore runc` should do the trick. For latter, if the checkpoint is GPU, it also spawns the GPU controller.

pkg/logging/logger.go
```diff
@@ -31,21 +32,45 @@ func init() {
...
 		logLevel = zerolog.Disabled
 	}
 
+	// "Writer" is just the the anonymous field "io.Writer" in the  struct
 	var output io.Writer = zerolog.ConsoleWriter{
 		Out:          os.Stdout,
 		TimeFormat:   LOG_TIME_FORMAT,
 		TimeLocation: time.Local,
 	}
+	consoleWriterLeveled := &LevelWriter{Writer: output, Level: logLevel}
```
yashanand1910, line 46: This is an unnecessary constraint. Daemon should simply output to stdout as before, and let the caller handle log storage.

swarnimarun:
@yashanand1910 did you test the dev shm with containers or just processes.
Processes are in the same namespace so they don't get affected. Containers require ipc host to allow the same shm mounts.

Will add the shim to registry file and see if it works and test it again to let you know.

yashanand1910:
> @yashanand1910 did you test the dev shm with containers or just processes. Processes are in the same namespace so they don't get affected. Containers require ipc host to allow the same shm mounts.
> 
> Will add the shim to registry file and see if it works and test it again to let you know.

@swarnimarun Of course, I am talking about runc C/R.

pkg/logging/logger.go
swarnimarun: This is just for having multiple log levels outputs one to stdout one to a file with debug/trace level logs.

This allows us to take a look at the logs while keeping pod logs clean, also added lumberjack so as to ensure the log file is rotated and doesn't ever get too big.

yashanand1910: systemd should take care of it, or whatever init service being used. For systemd, we can always use syslog to or `journalctl -u cedana.service` to fetch logs. This should not be baked into the binary.

swarnimarun:
closing in favour of $454

PR #448, opened 2025-01-16
chore: add k8s, etc plugins inside image

swarnimarun:
Install all plugins into the container image at the time of release by default, the images can be updated later.

Most importantly, this allows for k8s-helper code to work as it is.


yashanand1910 (approved):
Lgtm

yashanand1910:
K8s-setup-host.sh redownloads the latest versions. Should that be removed?

PR #450, opened 2025-01-16
hotfix: upgrade cedana k8s broken

swarnimarun:
cedana upgrade fails to copy cedana file, as it's running, keep this in case until we can provide a separate upgrade utility and binary

### DIFF
--- a/plugins/k8s/cmd/scripts/bump-restart.sh
+++ b/plugins/k8s/cmd/scripts/bump-restart.sh
@@ -1,10 +1,9 @@
 #!/bin/bash
+# NOTE: This script assumes it's executed in the container environment
 
 set -e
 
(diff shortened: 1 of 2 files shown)


plugins/k8s/cmd/scripts/bump-restart.sh
```diff
@@ -2,6 +2,8 @@
 
 set -e
 
+chroot /host /bin/bash /cedana/scripts/systemd-reset.sh
+
 # NOTE: This script assumes it's executed in the container environment
```
yashanand1910, line 7: This comment is for the script, can move it at the topmost. Good to merge after that.

yashanand1910 (approved):
Can merge after moving the comment.

PR #452, opened 2025-01-16
feat: Allow setting custom precision for printing profiling data

yashanand1910:
## Describe your changes
This is required by benchmark or other tools that want to read the data directly from stdout. Adds
the precision config item:
https://github.com/cedana/cedana/blob/2d924f8c81c8f7af8545587e3812d6793eeca95d/pkg/config/types.go#L57-L62

Default is "auto", which prints using the most suitable precision for the value.

### DIFF
--- a/cmd/run.go
+++ b/cmd/run.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"fmt"
 	"os"
-	"time"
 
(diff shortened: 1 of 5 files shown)


nravic (approved):
looks good!

PR #455, opened 2025-01-17
fix: GPU controller Attach called too quickly leading to race

yashanand1910:
## Describe your changes
This race is caused when a restore is called immediately after a dump. Since for dump, the request
is returned as soon as dump is completed, while in the background the daemon cleans up the GPU
controller. If a restore is called immediately after the dump, a GPU controller might still appear
as attached to the job.

This fix fails restore request that happens too quickly, instead of segfaulting on the internal controller
map.

### DIFF
--- a/internal/server/gpu/controller.go
+++ b/internal/server/gpu/controller.go
@@ -45,7 +45,7 @@ type controllers struct {
 	sync.Map
 }
 
-func (m *controllers) get(jid string) *controller {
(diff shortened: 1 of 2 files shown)


nravic (approved):
lgtm

swarnimarun:
We can just stall the restore can we not? Like pause the restore request until it is clear, if it's a race it should be a big delay even. Wdyt?

PR #460, opened 2025-01-19
fix: resolve full path in CLI for exec/run

yashanand1910:
## Describe your changes
Currently, when executing a process from the CLI, with `cedana exec/cedana run process`, passing in
the env vars to the daemon is not sufficient. The daemon is executing under a different env and so
the full path of the process must be provided to the daemon. For CLI, we simply resolve the full
path from where the command is called before passing it to the daemon.

### DIFF
--- a/cmd/run.go
+++ b/cmd/run.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"os"
+	"os/exec"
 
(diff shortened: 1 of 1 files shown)


swarnimarun (approved):
lgtm, Is this part of chdir changes?

yashanand1910:
It’s independent. I’ll mention that if its dependent on a another PR.

PR #461, opened 2025-01-19
fix: Get rid of chdir completely

yashanand1910:
## Describe your changes
We should never be `chdir`ing in the daemon, as the daemon is supposed to handle concurrent
requests and could lead to strange untraceable errors. This PR removes all instances of `chdir` in the daemon code.


plugins/runc/internal/container/run_adapters.go
```diff
@@ -18,21 +19,16 @@ func LoadSpecFromBundle(next types.Run) types.Run {
...
-		if bundle != "" {
-			cwd, err := os.Getwd()
-			if err != nil {
-				return nil, status.Errorf(codes.Internal, "failed to get current working directory: %v", err)
-			}
-			err = os.Chdir(bundle)
-			if err != nil {
-				return nil, status.Errorf(codes.Internal, "failed to set working directory: %v", err)
-			}
-			defer os.Chdir(cwd)
```
swarnimarun, line 32: lgtm, but defer would be applied to all functions so are we handling, are all paths absolute that we pass to runc for libcontainer stuff?

swarnimarun (approved):

plugins/runc/internal/container/run_adapters.go
yashanand1910: @swarnimarun Wdym by defer would be applied to all functions? Chdir would change the current directory of the entire process or thread, and not just this go routine thats handling the gRPC request.

PR #466, opened 2025-01-22
fix: follow symblinks for criu add external mounts

swarnimarun:
Fixes CED-909

I think there might be a bug in the criu external mounts code, we never use the mount Source only the Destination for criu add external mount which seems wrong. Need to dig into it a bit more to confirm.


swarnimarun:
Closing the scope of the issue, is larger than expected and would be better done as a refactor.

PR #468, opened 2025-01-23
feat: crio plugin port to plugin[wip]

swarnimarun:
WIP

CRIO Plugin, just has the code ported, needs some cleanup of rootfs logic and a bunch of testing, will add the test as well.

Push CRIO Image operation hasn't been implemented. (Will discuss with Brandon if it's needed)


plugins/crio/internal/client/dump_setup.go
```diff
@@ -0,0 +1,29 @@
...
+	"google.golang.org/grpc/status"
+)
+
+func SetupForDump(next types.Dump) types.Dump {
+	return func(ctx context.Context, opts types.Opts, resp *daemon.DumpResp, req *daemon.DumpReq) (exited chan int, err error) {
+		details := req.GetDetails().GetContainerd()
+
+		ctx = namespaces.WithNamespace(ctx, details.Namespace)
+
+		client, err := containerd.New(details.Address)
```
bsmithai, line 19: What is all this usage of containerd client inside of crio internal?

bsmithai (changes requested):

plugins/crio/internal/client/dump_setup.go
swarnimarun: That's the dump setup code from containerd, I need to clean it up, we don't seem to have a need for dump setup for CRIO so all that code just needs to go away.

Didn't push my latest changes sec.

PR #471, opened 2025-01-23
fix: GPU controller not being spawned as user

yashanand1910:
## Describe your changes
Spawns the GPU controller with same credentials as the user process being requested.
This fixes:
1. Issue with /dev/shm if the controller (that was being spawned as root) and the user process not able
to access it leading to a permission error.
2. A user process that's not supposed to have access to GPU was getting access because the
controller was always being spawned as root.


nravic (approved):
looks good

PR #475, opened 2025-01-26
fix: List/download only compatible plugins from registry

yashanand1910:
## Describe your changes
Updates the propagator plugin manager so that it only lists and downloads plugins that are
compatible with the current version of the binary. This is only needed for Go plugins, until we have
resolved the issues with cross-version compatibility. Other, external plugins like GPU, streamer are always
listed/downloaded latest. Even for external plugins, in the future, this would allow us to bind the binary to a compatible
version if there are breaking changes in the future.

Requires: https://github.com/cedana/cedana-propagator/pull/24

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -45,11 +45,11 @@ stop: ## Stop the daemon
 
 install-systemd: install ## Install the systemd daemon
 	@echo "Installing systemd service..."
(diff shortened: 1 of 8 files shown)


nravic (approved):
lgtm

PR #483, opened 2025-01-28
fix: CRIU logging verbosity

yashanand1910:
## Describe your changes
Noticed that the CRIU logging verbosity significantly slows down a dump, especially when dump a large container
with GPU support. This makes it so that CRIU will only log verbosely if Cedana's logging level is
set to debug/trace.


nravic (approved):
noice

PR #484, opened 2025-01-28
fix: External namespace file not being passed in context

yashanand1910:
## Describe your changes
If a container is using an external namespace, the namespace file needs to be opened by CRIU, so it
can `inherit-fd` it.

Also added tests for runc C/R, if a container has an external namespace or a combination of them.

### DIFF
--- a/plugins/runc/internal/namespace/restore_adapters.go
+++ b/plugins/runc/internal/namespace/restore_adapters.go
@@ -107,19 +107,21 @@ func InheritExternalNamespacesForRestore(nsTypes ...configs.NamespaceType) types
 				if err != nil {
 					return nil, status.Errorf(codes.Internal, "external namespace file %s does not exist: %v", nsPath, err)
 				}
+				defer nsFd.Close()
 				extraFiles = append(extraFiles, nsFd)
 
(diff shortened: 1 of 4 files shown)


plugins/runc/internal/namespace/restore_adapters.go
```diff
@@ -107,19 +107,21 @@ func InheritExternalNamespacesForRestore(nsTypes ...configs.NamespaceType) types
 				if err != nil {
 					return nil, status.Errorf(codes.Internal, "external namespace file %s does not exist: %v", nsPath, err)
 				}
+				defer nsFd.Close()
```
swarnimarun, line 110: Should we close this file before next? Does keeping it open during next affect anything?

swarnimarun (approved):
lgtm, just take a look at if the file close should happen early.

plugins/runc/internal/namespace/restore_adapters.go
yashanand1910: It needs to be open, so that when CRIU is spawned later, it's also has this file opened. Just tried closing it before, and our tests fail.

PR #485, opened 2025-01-28
feat: use ENV to initialize better defaults for new k8s installs

swarnimarun:
Currently, when the first daemon start happens if the env CEDANA_REMOTE is set and loaded for the configuration generation during daemon start.

This ensures K8S defaults to CEDANA_REMOTE="true"


yashanand1910 (approved):
Looks good

pkg/config/types.go
```diff
@@ -9,9 +9,9 @@ type (
 	// alternative (alias) environment variable names (comma-separated).
 	Config struct {
 		// Address to use for incoming/outgoing connections
-		Address string `json:"address" key:"address" yaml:"address" mapstructure:"address"`
+		Address string `json:"address" key:"address" yaml:"address" mapstructure:"address" env_aliases:"CEDANA_ADDRESS"`
```
yashanand1910, line 12: You don't need to add these aliases. They are set by default.

pkg/config/types.go
```diff
@@ -9,9 +9,9 @@ type (
...
 		// Protocol to use for incoming/outgoing connections (TCP, UNIX, VSOCK)
-		Protocol string `json:"protocol" key:"protocol" yaml:"protocol" mapstructure:"protocol"`
+		Protocol string `json:"protocol" key:"protocol" yaml:"protocol" mapstructure:"protocol" env_aliases:"CEDANA_PROTOCOL"`
```
yashanand1910, line 14: Same here, the default env var for this is already CEDANA_PROTOCOL.

pkg/config/types.go
```diff
@@ -70,7 +70,7 @@ type (
 
 	CLI struct {
 		// Wait for ready sets CLI commands to block if the daemon is not up yet
-		WaitForReady bool `json:"wait_for_ready" key:"wait_for_ready" yaml:"wait_for_ready" mapstructure:"wait_for_ready"`
+		WaitForReady bool `json:"wait_for_ready" key:"wait_for_ready" yaml:"wait_for_ready" mapstructure:"wait_for_ready" env_aliases:"CEDANA_WAIT_FOR_READY"`
```
yashanand1910, line 73: This is fine, the default would've been CEDANA_CLI_...

plugins/k8s/cmd/helper.go
```diff
@@ -222,7 +224,12 @@ func isDaemonRunning(ctx context.Context, address, protocol string) (bool, error
 		return false, err
 	}
 	defer client.Close()
-	return client.HealthCheckConnection(ctx)
+	_, err = client.HealthCheck(ctx, &daemon.HealthCheckReq{Full: true}, grpc.WaitForReady(true))
+	if err != nil {
```
yashanand1910, line 228: Error here only means the health check gRPC call failed. You need an additional check after this. Something like:
```
	// Wait for the daemon to be ready, and do health check
	resp, err := client.HealthCheck(ctx, &cedanadaemon.HealthCheckReq{Full: false}, grpc.WaitForReady(true))
	if err != nil {
		return nil, fmt.Errorf("cedana health check failed: %w", err)
	}

	errorsFound := false
	for _, result := range resp.Results {
		for _, component := range result.Components {
			for _, errs := range component.Errors {
				log.Error().Str("name", component.Name).Str("data", component.Data).Msgf("cedana health check error: %v", errs)
				errorsFound = true
			}
			for _, warning := range component.Warnings {
				log.Warn().Str("name", component.Name).Str("data", component.Data).Msgf("cedana health check warning: %v", warning)
			}
		}
	}
	if errorsFound {
		return nil, fmt.Errorf("cedana health check failed")
	}
```
Printing warnings and errors that were found.

pkg/config/types.go
```diff
@@ -70,7 +70,7 @@ type (
 
 	CLI struct {
```
yashanand1910, line 71: Can you just also, change this to 'Client'? I think CLI is an outdated name for us.

yashanand1910 (dismissed):

PR #492, opened 2025-01-31
fix: runc run race condition on container creation

yashanand1910:
## Describe your changes
Since `cmd.Start` starts the process in the background, we need to wait for sometime before the
container is actually created, before proceeding. Similarly, if the cmd.Start fails, we need to
wait for some time until the process has actually exited before returning an error.


nravic (commented):
think there's a typo, looks good otherwise

PR #495, opened 2025-02-10
fix: Don't install unsupported plugin

yashanand1910:
## Describe your changes
`cedana plugin list` would show this following warning if a plugin a supported plugin is not
available for the current binary version. E.g. here the k8s plugin is not available:
image

However, when you would do `cedana install k8s` it would still download the latest one and install
it. This PR fixes it.


nravic (approved):
lgtm

PR #501, opened 2025-02-26
ci: Reuse regression tests in `cedana-gpu` & `streamer`

yashanand1910:
## Describe your changes
The regression test suite in this repo can be reused in `cedana-gpu` and `cedana-image-streamer` repos, so that we only ever need to add new tests in this repo and it will automatically be run on other repos too.

This is good because we can keep expanding our test suite in a single place, given that now they can easily be run locally (using `make test`, etc.).

Accompanying PRs:
1. cedana-gpu: https://github.com/cedana/cedana-gpu/pull/125
2. cedana-image-streamer: https://github.com/cedana/cedana-image-streamer/pull/16

Both above repos reuse the PR workflow from this repo. For e.g., for `cedana-gpu`, it looks like this:
```yml
  daemon-tests:
    name: Daemon
    uses: cedana/cedana/.github/workflows/pr.yml
    needs: build
    permissions:
      contents: "read"
      packages: "read"
      id-token: "write"
      pull-requests: "write"
    with:
      ref: main
      debug_build: ${{ startsWith(github.event_name, 'workflow') && inputs.debug_daemon_tests }}
      skip_bench: true
      skip_shellcheck: true
      skip_helper_image_push: true
      skip_cpu_tests: true
      skip_streamer_tests: true
      skip_gpu_tests: false
    secrets: inherit
```
It skips all the tests except the GPU ones. Now, when working on `cedana-gpu`, if you want to test against a separate branch of the daemon, just change `ref` to the branch name.


nravic (approved):
*chef's-kiss*

wfoy (approved):

PR #502, opened 2025-02-26
Feat/ced 997 vm snapshot restore

bsmithai:

### DIFF
--- a/scripts/host/k8s-install-plugins.sh
+++ b/scripts/host/k8s-install-plugins.sh
@@ -66,5 +66,5 @@ else
 END_CAT
     fi
     echo "Sending SIGHUP to containerd..."
-    systemctl restart containerd
+    (systemctl restart containerd && echo "Restarted containerd") || echo "Failed to restart containerd, please restart containerd on the node manually to add cedana runtime"
 fi


cmd/dump.go
```diff
@@ -77,6 +77,83 @@ func init() {
 			return nil
 		},
 	)
+
+	features.DumpVMCmd.IfAvailable(
```
yashanand1910, line 81: DumpVM being a separate subcommand can be put in it's own file `dump_vm.go`.

internal/server/dump.go
```diff
@@ -21,6 +22,39 @@ import (
 	"google.golang.org/grpc/status"
 )
 
+func (s *Server) DumpVM(ctx context.Context, req *daemon.DumpVMReq) (*daemon.DumpVMResp, error) {
```
yashanand1910, line 25: DumpVM being a separate RPC handler, can be in it's own file `dump_vm.go`.

internal/server/dump.go
```diff
@@ -80,6 +114,31 @@ func (s *Server) Dump(ctx context.Context, req *daemon.DumpReq) (*daemon.DumpRes
 //// Helper Adapters /////
 //////////////////////////
 
+// Adapter that inserts new adapters after itself based on the type of dump.
+func pluginDumpVMMiddleware(next types.DumpVM) types.DumpVM {
+	return func(ctx context.Context, opts types.Opts, resp *daemon.DumpVMResp, req *daemon.DumpVMReq) (exited chan int, err error) {
+		middleware := types.Middleware[types.DumpVM]{}
+		t := req.GetType()
+		switch t {
+		case "cloud-hypervisor":
```
yashanand1910, line 123: Instead of directly supporting 'cloud-hypervisor' in the daemon, I'd create a plugin for it. Daemon does not need to support any VMs by default, I believe.

And so, the dump handler that is called by the above `DumpVM` will also be provided by the plugin. Each VM's plugin can implement it's own handler.

internal/server/filesystem/dump_adapters.go
```diff
@@ -124,3 +125,69 @@ func PrepareDumpDir(next types.Dump) types.Dump {
...
+		if err != nil {
+			return nil, status.Errorf(codes.Internal, "failed to open dump dir: %v", err)
+		}
+		defer f.Close()
+
+		// Setup dump fs that can be used by future adapters to directly read write/extra files
+		// to the dump directory
+		opts.DumpFs = afero.NewBasePathFs(afero.NewOsFs(), dir)
+
+		if req.Type == "cloud-hypervisor" {
```
yashanand1910, line 157: This logic can thus end up in the CLH plugin as a new adapter that simply modifies the `req.Dir`.

internal/server/vm/dump_handlers.go
yashanand1910, line 1: This file won't be needed as each VM plugin will have to implement it's own dump handler. That way, we can also remove the interface.

internal/server/vm/generic_adapters.go
yashanand1910, line 1: This will probably not be needed either, since the dump/restore handlers for each VM plugin will do it it's own way. Unless, we are trying to instantiate something that will be used by all plugins.

pkg/clh/clh.go
yashanand1910, line 1: This entire package can live inside the CLH plugin's source code (in its own `/pkg`).

pkg/features/supported.go
```diff
@@ -24,6 +25,7 @@ var (
 
 	// Dump/Restore
 	DumpMiddleware    = plugins.Feature[types.Middleware[types.Dump]]{Symbol: "DumpMiddleware", Description: "Dump middleware"}
+	DumpVMMiddleware  = plugins.Feature[types.Middleware[types.DumpVM]]{Symbol: "DumpVMMiddleware", Description: "Dump VM middleware"}
 	RestoreMiddleware = plugins.Feature[types.Middleware[types.Restore]]{Symbol: "RestoreMiddleware", Description: "Restore middleware"}
```
yashanand1910, line 29: Will have to add `DumpVMHandler` and same for restore as each VM plugin will implements its own.

yashanand1910 (changes requested):

internal/server/server.go
```diff
@@ -33,6 +33,9 @@ type Server struct {
 	gpus    gpu.Manager
 	plugins plugins.Manager
 	db      db.DB
+	// fdStore stores a map of fds used for clh kata restores to persist network fds and send them
+	// to the appropriate clh vm api
+	fdStore sync.Map
```
yashanand1910, line 38: What’s this doing here? Do we have an fdStore for the entire server instance?

plugins/cloud-hypervisor/cmd/cmd.go
```diff
@@ -0,0 +1,39 @@
...
+
+	"buf.build/gen/go/cedana/cedana/protocolbuffers/go/daemon"
+	"buf.build/gen/go/cedana/cedana/protocolbuffers/go/plugins/kata"
+	"github.com/cedana/cedana/pkg/keys"
+	clh_flags "github.com/cedana/cedana/plugins/cloud-hypervisor/pkg/flags"
+	"github.com/spf13/cobra"
+)
+
+var DumpCmd = &cobra.Command{
+	Use:   "cloud-hypervisor <vm-id>",
```
yashanand1910, line 15: Can we rename the plugin and subcommand to `clh`? Just a UX suggestion.

plugins/cloud-hypervisor/internal/handlers/dump_handlers.go
```diff
@@ -0,0 +1,57 @@
...
+	"github.com/cedana/cedana/pkg/types"
+	"github.com/cedana/cedana/pkg/utils"
+	clh "github.com/cedana/cedana/plugins/cloud-hypervisor/pkg/clh"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+)
+
+var Dump types.DumpVM = dump
+
+type VMSnapshotter interface {
```
yashanand1910, line 16: I’d put this somewhere in `pkg`, with CLH being it’s implementation. `pkg/vm/snaphotter.go` can container just the interface. `pkg/vm/snapshotter_clh.go` contains the CLH implementation. Since it’d be in a package called `vm`, the interface can simply be called `Snapshotter`, so that wherever it’s imported, it’s imported as `vm.Snapshotter`, for simplicity.

yashanand1910 (changes requested):
Looking great overall! Please take a look at my comments

internal/server/server.go
bsmithai: This is where I needed your input, in the old daemon I did the same thing, used the service struct. It's just that since fds are being sent over a unix socket, I just need a why to reference a fd from a key value store and I liked it being in mem, but if not, I can do it else where like sqlite store or something. Just need some sort of singleton that will hold fds

nravic (dismissed):
lgtm!

internal/server/filesystem/restore_adapters.go
yashanand1910, line 1: The changes here should instead be part of `restore_vm_adapters.go` file. Remember all the adapters are segregated by type?

pkg/features/supported.go
```diff
@@ -16,16 +16,21 @@ var (
...
 	// Dump/Restore
-	DumpMiddleware    = plugins.Feature[types.Middleware[types.Dump]]{Symbol: "DumpMiddleware", Description: "Dump middleware"}
-	RestoreMiddleware = plugins.Feature[types.Middleware[types.Restore]]{Symbol: "RestoreMiddleware", Description: "Restore middleware"}
-	KillSignal        = plugins.Feature[syscall.Signal]{Symbol: "KillSignal", Description: "Custom kill signal"}
+	DumpMiddleware      = plugins.Feature[types.Middleware[types.Dump]]{Symbol: "DumpMiddleware", Description: "Dump middleware"}
+	DumpVMMiddleware    = plugins.Feature[types.Middleware[types.DumpVM]]{Symbol: "DumpVMMiddleware", Description: "Dump VM middleware"}
+	RestoreVMMiddleware = plugins.Feature[types.Middleware[types.RestoreVM]]{Symbol: "RestoreVMMiddleware", Description: "Restore VM middleware"}
+	RestoreMiddleware   = plugins.Feature[types.Middleware[types.Restore]]{Symbol: "RestoreMiddleware", Description: "Restore middleware"}
+	KillSignal          = plugins.Feature[syscall.Signal]{Symbol: "KillSignal", Description: "Custom kill signal"}
+	DumpVMHandler       = plugins.Feature[types.DumpVM]{Symbol: "DumpVMHandler", Description: "DumpVM handler"}
```
yashanand1910, line 32: Can we also ensure all of these are present in `cmd/plugin.go` for when `cedana plugin features` is invoked from CLI?

yashanand1910 (changes requested):
Check my comments, just some minor changes.

Also, `.github/workflows/release.yml` needs to be updated to include the new plugin binaries. Should be simple, take a look.

nravic (approved):
looks good to me, great work!

PR #506, opened 2025-03-07
Feat: Implement RabbitMQ metrics stream

TusharMohapatra07:
## Describe your changes
This PR refactors the initialization of ASR (Availability, Scalability, Reliability) metrics in the Cedana daemon. 


pkg/metrics/rabbit.go
yashanand1910, line 1: This file can instead be put in `internal/metrics`. This will likely contain implementation specific stuff once we start using it properly. `pkg` is only for stuff that we wish to export.

cmd/daemon.go
```diff
@@ -49,11 +51,20 @@ var startDaemonCmd = &cobra.Command{
...
 
-		var err error
+		ctx, cancel := context.WithCancel(cmd.Context())
+		defer cancel()
 
 		log.Info().Str("version", rootCmd.Version).Msg("starting daemon")
 
-		server, err := server.NewServer(cmd.Context(), &server.ServeOpts{
+		// Initialize metrics reporting if ASR flag is enabled
+		if viper.GetBool("metrics.asr") {
```
yashanand1910, line 60: Take a look at how we reference config items at other places. We don't use `viper.Get...` anymore. You can directly reference the `config.Global` definition as it's always updated with the right values.

cmd/daemon.go
```diff
@@ -49,11 +51,20 @@ var startDaemonCmd = &cobra.Command{
...
+			cedanaURL := viper.GetString("connection.url")
+			if err := metrics.PollAndPublishMetrics(ctx, cedanaURL); err != nil {
```
yashanand1910, line 62: I'd also move this initialization inside `server.Launch` itself. We initialize Otel stuff too in there, for example.

yashanand1910 (changes requested):

TusharMohapatra07:
@yashanand1910 PTAL. Let me know if there are any changes that's needs to be made.

yashanand1910 (approved):

PR #511, opened 2025-03-21
feat:  shim installation

swarnimarun:
shim installation.

fixes CED-1058

PR #512, opened 2025-03-24
feat: add runc checkpoint path to containerd image (wip)

swarnimarun:
@rahulk789 this branch should allow you to perform, runc + rootfs checkpointing and restores with the new daemon and shim.

Note: this will require https://github.com/cedana/cedana/pull/511 to be able to be used in production clusters. Also, we would need to sync the checkpoint amongst nodes, in the /tmp path or where-ever you want to read the checkpoint from at the time of restore.

TODO(why wip):

- Currently, the checkpoint path for restore is hardcoded within the env for the container we build we need to provide better means of handling it. Also, we need to be able to push the image to a registry, for which the logic isn't in this PR.


yashanand1910 (approved):
Approved, for now.

PR #519, opened 2025-03-31
fix: dump dir flag not being recognized

yashanand1910:
## Describe your changes
The dump `--dir` flag was not being recognized due to a bug (or feature?) in cobra/viper, when a
flag is bound to a config item more than once. For now, I've removed most of the flag bindings that
can potentially cause these failures, given that we have logic in the daemon already to read from
config when necessary.


nravic (approved):
looks good!

PR #521, opened 2025-04-04
hotfix: on cancel we should still resume pod

swarnimarun:
We missed this, defer can run on context cancelled, but then context.Background is needed otherwise Resume won't happen.

This can happen often as containerd dump is very long.

I should also add a containerd test for this. 

Fixes CED-1093

### DIFF
--- a/plugins/containerd/internal/filesystem/dump_adapters.go
+++ b/plugins/containerd/internal/filesystem/dump_adapters.go
@@ -57,7 +57,7 @@ func DumpRootfs(next types.Dump) types.Dump {
 			return nil, status.Errorf(codes.Internal, "failed to pause container %s: %v", details.ID, err)
 		}
 		defer func() {
-			err := task.Resume(ctx)
+			err := task.Resume(context.Background())
 			if err != nil {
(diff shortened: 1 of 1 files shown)


nravic (approved):
lgtm!

yashanand1910 (changes requested):
Instead use `context.WithoutCancel(ctx)`. Should avoid background at all times. The above code ensures, if at all any contextual value is part of `ctx`, it will be preserved.

plugins/containerd/internal/filesystem/dump_adapters.go
```diff
@@ -57,7 +57,7 @@ func DumpRootfs(next types.Dump) types.Dump {
 			return nil, status.Errorf(codes.Internal, "failed to pause container %s: %v", details.ID, err)
 		}
 		defer func() {
-			err := task.Resume(ctx)
+			err := task.Resume(context.Background())
```
swarnimarun, line 60:
```suggestion
			err := task.Resume(context.WithoutCancel(ctx))
```

nravic (approved):
lgtm

PR #526, opened 2025-04-15
feat: remove dependency for restores on daemon server

swarnimarun:
- Adds a flag --no-server to remove use of daemon server.
- Adds pidfile path writing to allow using as a runc alternative. 
- (minor) Also changes rand to https://go.dev/blog/randv2
- Adds test to validate the newly added api


internal/server/server.go
```diff
@@ -30,23 +30,33 @@ type Server struct {
...
-	wg       *sync.WaitGroup // for waiting for all background tasks to finish
 	lifetime context.Context // context alive for the duration of the server
+	wg       *sync.WaitGroup // for waiting for all background tasks to finish
+	plugins  plugins.Manager
 
 	host    *daemon.Host
 	version string
 
 	daemongrpc.UnimplementedDaemonServer
 }
```
nravic, line 50: Can you add a nice docstring or lengthy comment here explaining: 
- The purpose of CedanaRoot
- How it differs from regular daemon functionality?

internal/server/restore.go
```diff
@@ -72,6 +72,53 @@ func (s *Server) Restore(ctx context.Context, req *daemon.RestoreReq) (*daemon.R
 	return resp, nil
 }
 
+func (s *CedanaRoot) Restore(ctx context.Context, req *daemon.RestoreReq) (*daemon.RestoreResp, error) {
```
nravic, line 75: same thing here re: a note about how CedanaRoot.Restore differs from Server.Restore (and why). You can just ref the other comment

cmd/restore.go
```diff
@@ -26,6 +27,11 @@ func init() {
 	restoreCmd.AddCommand(processRestoreCmd)
 	restoreCmd.AddCommand(jobRestoreCmd)
 
+	// Select what api is used for restores
+	restoreCmd.PersistentFlags().
+		StringP(flags.PidFileFlag.Full, flags.PidFileFlag.Short, "", "required for runc restores")
+	restoreCmd.PersistentFlags().
+		BoolP(flags.NoServerFlag.Full, flags.NoServerFlag.Short, false, "select how to run restores")
```
nravic, line 34: Will PIDFileFlag ever be used in a no-server context? If so, might be worth making them dependent with MarkFlagsRequiredTogether or something

cmd/restore.go
```diff
@@ -122,14 +137,16 @@ var restoreCmd = &cobra.Command{
...
 		ctx := context.WithValue(cmd.Context(), keys.RESTORE_REQ_CONTEXT_KEY, req)
-		cmd.SetContext(ctx)
-
-		client, err := client.New(config.Global.Address, config.Global.Protocol)
-		if err != nil {
-			return fmt.Errorf("Error creating client: %v", err)
+		if pidFilePath != "" {
+			ctx = context.WithValue(ctx, keys.PIDFILE_PATH_KEY, &pidFilePath)
+		}
+		if !noServer {
```
yashanand1910, line 143: NIT: Return early on noServer, can avoid double negative.

cmd/restore.go
```diff
@@ -142,17 +159,45 @@ var restoreCmd = &cobra.Command{
...
+			}
+
+			resp, err := cedanaRoot.Restore(ctx, req)
+			if err != nil {
+				return fmt.Errorf("cedana restore run err: %v", err)
+			}
+
+			pidFilePath, _ := cmd.Context().Value(keys.PIDFILE_PATH_KEY).(*string)
+			if req.GetDetails().GetRunc() != nil && *pidFilePath == "" {
+				log.Error().Msg("cedana runc restore without pidfile")
```
yashanand1910, line 183: Why is this an error log? Should be normal behaviour to not provide a PidFile.

internal/server/server.go
```diff
@@ -30,23 +30,34 @@ type Server struct {
...
 
+// CedanaRoot struct avoids the use of jobs and provides runc compatible cli usage
```
yashanand1910, line 51:
```suggestion
// CedanaRoot avoids the use of the server and provides direct run/restores
```

yashanand1910 (approved):
Lgtm. Check the comments, please.

cmd/restore.go
```diff
@@ -142,17 +159,45 @@ var restoreCmd = &cobra.Command{
...
-		}
-		defer client.Close()
-
 		// Assuming request is now ready to be sent to the server
 		req, ok := cmd.Context().Value(keys.RESTORE_REQ_CONTEXT_KEY).(*daemon.RestoreReq)
 		if !ok {
 			return fmt.Errorf("invalid restore request in context")
 		}
+		client, ok := cmd.Context().Value(keys.CLIENT_CONTEXT_KEY).(*client.Client)
+		if !ok {
```
yashanand1910, line 168: Please flip this if condition, keeping smaller code (for server-based restore) inside the if. This is a bit unintuitive.

yashanand1910 (commented):
Another NIT

cmd/restore.go
swarnimarun: Cases where we should be using this as runc, we would have use pidFile path. So generally it's a soft error for us if pidFile path is not present.

Flags are easy to miss, and it can lead to bugs if pidFile is not present. But we can make it a warn log... which might be better.

nravic (approved):
lgtm, great work!

PR #535, opened 2025-05-06
feat: gpu run daemonless

swarnimarun:
- Added `--no-server` run. Tested for process as well as runc plugin.
- Added `--upcoming` flag to `cedana manage` subcommand. Check new enum value `MANAGE_UPCOMING` if using gRPC. This flag allows us to call `cedana manage runc ...` even before the container has been created. This is required because the GPU controller **must** be spawned before the container is spawned, otherwise the `/dev/shm` mount will not be available during container creation.
- Added new tests for above.

#### How this will work
1. Start managing upcoming container (with GPU support):
```
cedana manage runc <JID> --upcoming --gpu-enabled
```
2. Spawn container with GPU support (daemonless):
```
sudo cedana run runc --bundle bundle --no-server <JID> --gpu-enabled
```

And that's it, it should work. And for restore, simply replace step 2 with `cedana restore runc ... --no-server` as the restored container will already have GPU interception.


cmd/run.go
```diff
@@ -96,30 +99,52 @@ var runCmd = &cobra.Command{
...
-			return client.Attach(cmd.Context(), &daemon.AttachReq{PID: resp.PID})
+		resp := &daemon.RunResp{}
+		if noServer {
+			client, err := client.New(config.Global.Address, config.Global.Protocol)
+			if err != nil {
+				return fmt.Errorf("Error creating client: %v", err)
+			}
+			defer client.Close()
+			// Assuming request is now ready to be sent to the server
+			resp, profiling, err := client.Run(cmd.Context(), req)
```
swarnimarun, line 117: woof bug

nravic (approved):
lgtm!

PR #546, opened 2025-05-28
Add runc GPU and remote C/R tests + several bug fixes

yashanand1910:
## Describe your changes

This PR fixes several bugs and adds a lot of new integration tests for runc + GPU C/R, including streaming and remote C/R (using the `storage/cedana` plugin).

### Bug fixes:
- Race related to preserving protobuf’s across requests. This was the primary cause for intermittent test failures.
- No-server run/restore I/O, signals & exit code for process and runc.
- Restoring console socket for runc, using `OrphanPTSMaster` CRIU hook
- Random 'Ghost file' CRIU failures when GPU C/R'ing
- Several other race conditions

### Improvements:
- Runc plugin now supports the `--detach` and `--console-socket` flags, improving support for the runtime shim for k8s. The `--detach` flag is only used when running using `--no-server`, otherwise the daemon manages the container and enables it by default.
- `cedana run runc --no-server` now behaves identical to `runc run`, including handling of I/O, signals, and exit code.
- GPU manager no longer uses a DB to manage GPU controllers. The PR https://github.com/cedana/cedana-gpu/pull/149 adds a new `Info` RPC method that is used by the daemon to stay in sync with running GPU controllers in the system. This fixes major issues when running GPU-enabled process/containers using `--no-server`, where the GPU manager would not be able to write to DB if not run as root. This also greatly simplifies the GPU manager code. Most of the `gpu.Manager` is implemented by `gpu.ManagerSimple`, and `gpu.ManagerPool` simply embeds it to provide pooling functionality. `gpu.ManagerSimple` is used for `--no-server` runs, while `gpu.ManagerPool` is used by the daemon. 
- The `cedana job kill`/`Kill` RPC for managed jobs now sends the signal to the process group, if the managed job is a process group leader.
- When a GPU controller needs to be terminated, the signal is sent to the process group, ensuring all it's workers are also terminated. Thus, GPU controllers are now spawned as process group leaders.
- Custom `Reaper` feature allows a plugin to handle reaping of restored proceses on its own. This is needed to ensure `cedana run runc --no-server` works correctly. Earlier, the restore handler would wait to reap processes indiscriminately.
- Removes the `UID`, `GID`, and `Groups` when spawning GPU controllers, since they are spawned in a pool and should be identical to each other and so they are always spawned with the daemon's credentials.
- Custom `Cleanup` feature for plugins allows a plugin to define a routine that will be called when a managed job gets killed. Before this, each plugin had to check whether the job is managed or not.
- Changed the return type of the `types.Handler` to: https://github.com/cedana/cedana/blob/a621515183a45d68fe2f6c15e2bcbf27133f8c2c/pkg/types/handler.go#L37-L37 Instead of directly returning a channel that's closed on job exit (no exit code), it now returns a function which when called returns a channel for the job's exit code. This is function because all the handlers now return a [channel.Broadcaster](https://github.com/cedana/cedana/blob/feature/ced-1206-add-runc-gpu-and-remote-cr-tests/pkg/channel/broadcaster.go): https://github.com/cedana/cedana/blob/a621515183a45d68fe2f6c15e2bcbf27133f8c2c/internal/cedana/process/run_handlers.go#L50 This a new primitive that can be used in the daemon to broadcast values to every receiver, vs. a plain channel that would only send it to the first receiver. This primitive should not be abused and only efficient for small data types as it needs to buffer values to replicate channel behaviour.

### New tests:
- storage/cedana remote C/R
- storage/cedana remote streaming C/R
- storage/cedana remote streaming GPU C/R
- runc GPU C/R
- runc GPU streaming C/R
- runc no-server run/restore tests

### Config updates:
- `LogLevelNoServer` can be set to alter the logging level of `--no-server` run/restores.
  https://github.com/cedana/cedana/blob/a621515183a45d68fe2f6c15e2bcbf27133f8c2c/pkg/config/types.go#L17-L18
- `GPU.FreezeType` has been added back to specify what freeze type to use when dumping GPU-enabled jobs.
  https://github.com/cedana/cedana/blob/a621515183a45d68fe2f6c15e2bcbf27133f8c2c/pkg/config/types.go#L97-L98
 

### DIFF
--- a/internal/cedana/criu/restore_handlers.go
+++ b/internal/cedana/criu/restore_handlers.go
@@ -145,7 +145,7 @@ func restore(ctx context.Context, opts types.Opts, resp *daemon.RestoreResp, req
 		}()
 	}
 
-	log.Debug().Int("CRIU", version).Msg("CRIU restore complete")
+	log.Debug().Int("CRIU", version).Uint32("PID", resp.PID).Msg("CRIU restore complete")
 
 	return code, err
 }
--- a/internal/cedana/criu/restore_handlers.go
+++ b/internal/cedana/criu/restore_handlers.go
@@ -45,7 +45,6 @@ func restore(ctx context.Context, opts types.Opts, resp *daemon.RestoreResp, req
 	criuOpts.LogFile = proto.String(CRIU_RESTORE_LOG_FILE)
 	criuOpts.LogLevel = proto.Int32(CRIU_LOG_VERBOSITY_LEVEL)
 	criuOpts.GhostLimit = proto.Uint32(GHOST_FILE_MAX_SIZE)
-	criuOpts.Pid = proto.Int32(int32(resp.GetState().GetPID()))
 	criuOpts.NotifyScripts = proto.Bool(true)
 	criuOpts.LogToStderr = proto.Bool(false)
 	criuOpts.RstSibling = proto.Bool(true) // always restore as a child of the daemon
--- a/.github/workflows/bench.yml
+++ b/.github/workflows/bench.yml
@@ -532,9 +532,9 @@ jobs:
           path: |
             /src/results
 
-  post_summary:
+  post-summary:
     name: Post Summary
-    if: always()
+    if: ${{ !cancelled() }}
     needs: [cedana_runtime_1, cedana_runtime_2, cedana_cr_1, cedana_cr_2, native]
     permissions:
       contents: "read"
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -44,7 +44,7 @@ jobs:
     name: Cedana
     runs-on: runs-on=${{github.run_id}}/runner=1cpu-4g-${{ matrix.arch }}/extras=s3-cache
     strategy:
-      fail-fast: false
+      fail-fast: true
       matrix:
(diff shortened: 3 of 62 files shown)


internal/cedana/criu/restore_handlers.go
```diff
@@ -47,6 +48,7 @@ func restore(ctx context.Context, opts types.Opts, resp *daemon.RestoreResp, req
 	criuOpts.Pid = proto.Int32(int32(resp.GetState().GetPID()))
 	criuOpts.NotifyScripts = proto.Bool(true)
 	criuOpts.LogToStderr = proto.Bool(false)
+	criuOpts.RstSibling = proto.Bool(true) // always restore as a child of the daemon
```
nravic, line 50: why add this now? If daemon dies for any reason, children do too no?

yashanand1910: They don't die, no. Unless we specify `Pdeathsig`, they just get orphaned.
The reason I added this is because it's actually better than the default,
where the CRIU process stays alive and waits for the restored task to exit
before exiting. See this: https://criu.org/Tree_after_restore
[…](https://github.com/cedana/cedana/pull/546#)

PR #586, opened 2025-08-07
CED-1425: feat: update tests for cluster registeration changes

swarnimarun:
**Refactor cluster management to use cluster ID instead of name**

- Change `list_clusters()` to `create_cluster()` with POST request
- Update helm install to use cluster ID parameter
- Modify `cluster_id()` to register clusters on-demand
- Fix test setup order to get cluster ID before helm install


nravic (approved):
lgtm, remember to revert pr.yml changes

PR #595, opened 2025-09-03
CED-1433: Move runc plugin out

nravic:


bsmithai (approved):

.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "plugins/runc"]
+	path = plugins/runc
+	url = git@github.com:cedana/cedana-runc.git
```
yashanand1910, line 3: Can we specify the commit hash here, so that when a `cedana-runc` PR is made, this is updated by the developer.

yashanand1910 (changes requested):
Looks good, small comment.

.gitmodules
nravic: yeah

nravic: looked - there's no way to specify a commit hash, just the branch. Commit hash info is stored in the repo tree. I can add a branch, which should be good enough for now

PR #596, opened 2025-09-10
CED-1459: Fix cgroups restore in k8s/unexpected container deaths

yashanand1910:
## Describe your changes
Fixes the issue with pods dying shortly after restore. CRIU restore, by default, attempts to restore cgroup properties. In K8s, during restore, the container simply needs to use the cgroups of the new pod instead, without making any changes. Any change causes containerd to kill and restart the container, which triggers a new restore, again touching the cgroups, leading to a crash loop.

Applying Cgroups to the CRIU process does not apply it to the restored process. CRIU’s cgroups are only used if the saved cgroup for a task matches its cgroup, which is never the case in Kubernetes, since we create a new pod (with new cgroups) for restoring the containers. Reference: https://criu.org/CGroups (very confusing).

All restore tests are now passing for K3s, EKS, and GKE.


bsmithai (approved):
nice

nravic (approved):

PR #598, opened 2025-09-10
ced-1504: feat: pvc migration tests, tamarind example workload test (cgroup issue)

bsmithai:

### DIFF
--- a/test/helpers/propagator.bash
+++ b/test/helpers/propagator.bash
@@ -31,29 +31,32 @@ PROPAGATOR_AUTH_TOKEN="${CEDANA_AUTH_TOKEN}"
 # @param $1: Pod name
 # @param $2: Runc root path
 # @param $3: Namespace
-# @param $4: Cluster ID
+# @param $4: Pod ID (UID)
 # Returns: Action ID for polling
 checkpoint_pod() {
     local pod_name="$1"
     local runc_root="$2"
     local namespace="$3"
+    local pod_id="$4"
 
-    if [ -z "$pod_name" ] || [ -z "$runc_root" ] || [ -z "$namespace" ]; then
-        error_log "checkpoint_pod requires pod_name, runc_root, namespace"
+    if [ -z "$pod_name" ] || [ -z "$runc_root" ] || [ -z "$namespace" ] || [ -z "$pod_id" ]; then
+        error_log "checkpoint_pod requires pod_name, runc_root, namespace, pod_id"
         return 1
     fi
 
-    debug_log "Checkpointing pod '$pod_name' in namespace '$namespace'..."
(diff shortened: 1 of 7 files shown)


test/helpers/propagator.bash
```diff
@@ -31,29 +31,32 @@ PROPAGATOR_AUTH_TOKEN="${CEDANA_AUTH_TOKEN}"
 # @param $1: Pod name
 # @param $2: Runc root path
 # @param $3: Namespace
-# @param $4: Cluster ID
+# @param $4: Pod ID (UID)
```
yashanand1910: Don’t think this is needed. The propagator finds the right pod if either [name, namespace] or [pod_id] is specified.

test/k8s/eks.bats
```diff
@@ -209,3 +211,183 @@ teardown() {
...
+# bats test_tags=restore,pvc
+@test "Checkpoint and restore pod with PVC (wait until running, streams=$CEDANA_CHECKPOINT_STREAMS)" {
+
+    local pv_name="counting-pv"
+    local pvc_name="counting-pvc"
+    local pod_name="counting-pvc-pod"
+
+    kubectl apply -f "$WORKLOADS"/counting-pvc.yaml -n "$NAMESPACE"
+
+    echo "[DEBUG] Checking PV status..."
```
yashanand1910: Can you remove these extraneous logs? Bats completely ignores stdout. Instead use the `debug` / `debug_log` function defined in the utils.

test/k8s/eks.bats
```diff
@@ -209,3 +211,183 @@ teardown() {
...
+            fi
+        fi
+    fi
+
+    kubectl delete pvc "$pvc_name" -n "$NAMESPACE" --wait=true || true
+    kubectl delete pv "$pv_name" --wait=true || true
+}
+
+# bats test_tags=k8s,kubernetes,aws,AWS,eks,EKS,job,checkpoint,restore
+@test "Checkpoint and restore job with init container and volume (Tamarind example) (streams=$CEDANA_CHECKPOINT_STREAMS)" {
```
yashanand1910: You can remove this test. We won’t be supporting jobs directly.

test/workloads/simple-test-job2.yaml
```diff
@@ -0,0 +1,42 @@
+apiVersion: batch/v1
```
yashanand1910: Can remove this workload.

test/helpers/propagator.bash
bsmithai: There were left over pods with the same name and namespace which caused issues with the test. I included it so that the test is more usable in environments where we want to debug more easily as it is more precise to include the pod_id

yashanand1910 (approved):

PR #601, opened 2025-09-17
CED-1515: feat: write container name for checkpoints

swarnimarun:


yashanand1910 (changes requested):
Do not merge. We have an issue, as we cannot use Pod IDs in shim because they change every time.

PR #611, opened 2025-10-06
Add SLURM cgroups as a Cedana plugin

jgornet:
## Describe your changes
Most SLURM implementations use cgroups to provision resources. The cedana `plugins/slurm` module handles cgroup management (freeze/unfreeze) `internal/cgroup` as well as SLURM job management (such as checkpointing the job script) `internal/job`. The pull request as makes major changes to `cedana/cedana-api` to add the SLURM to `daemon.Details` and `cedana/cedana-slurm` to integrate the new Cedana daemon functionality.

Other pull requests
- https://github.com/cedana/cedana-slurm/pull/6
- https://github.com/cedana/cedana-api/pull/73


pkg/plugins/registry.go
```diff
@@ -89,7 +89,8 @@ var Registry = []Plugin{
 	},
 	{
 		Name:      "slurm",
-		Type:      EXTERNAL,
-		Libraries: []Binary{{Name: "libcedana-slurm.so", InstallDir: "/usr/lib64/slurm"}},
+		Type:      SUPPORTED,
+		Libraries: []Binary{{Name: "libcedana-slurm.so"}},
+		Binaries:  []Binary{{Name: "libslurm-cedana.so", InstallDir: "/usr/lib64/slurm"}},
```
nravic, line 99: slurm-cedana-plugin.so?

yashanand1910: I think let’s keep the name. Note that this is always installed in a slurm-specific directory, so there should be no confusion.

plugins/slurm/cmd/dump.go
```diff
@@ -0,0 +1,89 @@
...
+
+		if err := dump(id); err != nil {
+			return fmt.Errorf("failed to dump slurm job: %v", err)
+		}
+
+		return nil
+	},
+}
+
+func dump(jobid string) error {
```
yashanand1910: Why is this part of the `cmd` package? This looks like additional work that needs to be done as part of every slurm dump, and thus, it must be a dump adapter, so that even gRPC clients can get this behavior.

plugins/slurm/cmd/dump.go
```diff
@@ -0,0 +1,89 @@
...
+	cedanaURL := os.Getenv("CEDANA_URL")
+	if cedanaURL != "" {
+		return fmt.Errorf("CEDANA_URL environment variable is not set")
+	}
+
+	authToken := os.Getenv("CEDANA_AUTH_TOKEN")
```
yashanand1910: Shouldn’t use `os.Getenv` for any config value. You can directly use `config.Global.Connection.AuthToken`, which can be assumed to have the right value (either from the ~/.cedana/config.json or the environment variable).

See https://docs.cedana.ai/daemon/get-started/configuration.

plugins/slurm/internal/defaults/dump_adapters.go
```diff
@@ -0,0 +1,28 @@
+package defaults
+
+import (
+	"context"
+
+	"buf.build/gen/go/cedana/cedana/protocolbuffers/go/daemon"
+	criu_proto "buf.build/gen/go/cedana/criu/protocolbuffers/go/criu"
+	"github.com/cedana/cedana/gen/plugins/slurm"
```
yashanand1910: Where does this come from? Building locally, this cannot be found.

Can you simply do a `go get` to fetch the newly generated module from buf.build for your `cedana-api` branch? 

Link: https://buf.build/cedana/cedana/sdks/feature/ced-1524-commit-a-slurm-interface-for-the-cedana-propagator:protocolbuffers/go?version=v1.36.6

Let me know if you can’t access it.

plugins/slurm/internal/job/dump_adapters.go
```diff
@@ -0,0 +1,49 @@
...
+	"github.com/rs/zerolog/log"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+)
+
+func GetSlurmJobForDump(next types.Dump) types.Dump {
+	return func(ctx context.Context, opts types.Opts, resp *daemon.DumpResp, req *daemon.DumpReq) (code func() <-chan int, err error) {
+		jid := req.GetDetails().GetSlurm().GetJobID()
+		hostname := req.GetDetails().GetSlurm().GetHostname()
+		path := fmt.Sprintf("/system.slice/%s_slurmstepd.scope/job_%d/step_batch_user/task_special", hostname, jid)
```
yashanand1910: Does slurm always use systemd cgroups?

plugins/slurm/internal/job/dump_adapters.go
```diff
@@ -0,0 +1,49 @@
...
+	"github.com/cedana/cedana/pkg/types"
+	slurm_keys "github.com/cedana/cedana/plugins/slurm/pkg/keys"
+	"github.com/opencontainers/cgroups"
+	cgroupsManager "github.com/opencontainers/cgroups/manager"
+	"github.com/rs/zerolog/log"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+)
+
+func GetSlurmJobForDump(next types.Dump) types.Dump {
```
yashanand1910, line 17: Confused about why this is called `GetSlurmJobForDump`.

plugins/slurm/internal/job/dump_adapters.go
```diff
@@ -0,0 +1,49 @@
...
+		return nil, nil
+	}
+}
+
+func SetPIDForDump(next types.Dump) types.Dump {
+	return func(ctx context.Context, opts types.Opts, resp *daemon.DumpResp, req *daemon.DumpReq) (code func() <-chan int, err error) {
+		if resp.State == nil {
+			resp.State = &daemon.ProcessState{}
+		}
+		resp.State.PID = int32(req.GetDetails().GetSlurm().GetPID())
```
yashanand1910: Where is this set? If this is always required, should add a check for this in `ValidateDumpRequest`. Can the PID not be inferred from the other details instead?

yashanand1910: The fact that we are setting the PID directly is slightly awkward. The plugin should take in minimum details which are specific to the type of dump. For e.g., the `runc` plugin requires only the container ID and root, and the PID is inferred and thus set.

plugins/slurm/main.go
```diff
@@ -0,0 +1,68 @@
...
+// 	List: []types.Check{},
+// }
+
+// TODO: implement query
+// var QueryHandler types.Query = slurm.Query
+
+var (
+	KillSignal = syscall.SIGKILL
+	// Cleanup    = slurm.Cleanup // TODO: maybe implement cleanup if slurm doesn't handle it properly
+	Reaper = true // we handle our reaping on our own
```
yashanand1910: Why is this set?

yashanand1910 (changes requested):
Please check my comments. Thanks!

pkg/plugins/registry.go
jgornet: Thanks, I changed the binary name back to `libslurm-cedana.so`.

plugins/slurm/internal/defaults/dump_adapters.go
jgornet: Thanks!

plugins/slurm/internal/job/dump_adapters.go
jgornet: Going through the `SchedMD/slurm` codebase, it looks like SLURM only uses systemd cgroups. I can add support for other cgroup configurations if necessary.

plugins/slurm/internal/job/dump_adapters.go
jgornet: The `GetSlurmJobForDump` middleware retrieves all of the SLURM job information for the dump. In the future, this middleware call may retrieve and configure other SLURM properties (dump container details if using Podman or Singularity).

plugins/slurm/internal/job/dump_adapters.go
jgornet: Thanks, let me figure out a better way to get the PID. The `slurm` plugin is a little different from the `runc` plugin because the `cedana/cedana-slurm` SLURM plugin sends all of the job details to the `cedana/cedana` SLURM plugin. Unlike `runc`, SLURM does not have a stable library like `github.com/opencontainers/runc/libcontainer`, so I have to get all job details within a SLURM job's process—as opposed to `libcontainer.Load(root, id)` in `runc`.

plugins/slurm/main.go
jgornet: Sorry, I forgot to update the `main.go` staging.

plugins/slurm/internal/job/dump_adapters.go
yashanand1910: Makes sense.

plugins/slurm/internal/job/dump_adapters.go
yashanand1910: Got it!

plugins/slurm/internal/job/dump_adapters.go
yashanand1910: Cool.

plugins/slurm/cmd/dump.go
yashanand1910: Why was this resolved?

yashanand1910: Should take a look at all all `os.Getenv` instances.

plugins/slurm/cmd/dump.go
yashanand1910: Why was this resolved?

jgornet: I removed these files since it performs unexpected behavior for dumping.

plugins/slurm/cmd/dump.go
jgornet: Similar to the above comment, I remove this file since it performs unexpected behavior, so everything gets handled on the `cedana-slurm` side.

PR #709, opened 2026-01-27
Feat: Move from Cloudsmith->S3

rahulk789:
### Migrate plugin artifacts storage from Cloudsmith to AWS S3
This PR replaces Cloudsmith artifact uploads with  S3 uploads through propagator across all CI/CD workflows.
### What changed:
All download and publish jobs (cedana, plugins, CRIU, GPU binaries, image streamer, containerd runtime shim) have been updated to use v2/plugins/download and v2/plugins/upload instead of Cloudsmith based uploads. The cloudsmith and cloudsmith_repo workflow inputs have been renamed to aws_s3 and aws_s3_path respectively. 
- We receive a presigned url for the given version and latest version paths which are then used to upload the respective binaries during the publish phase.
- We loop through several binaries in a single job instead of our old matrix setup for the plugin uploads. requests are async and parallel so  the entire publish takes less than 10 seconds
- Retries are added in case any of the uploads fail. Ideally we will never hit them
-  We specify the respective binary in the propagator query during downloads as well along with requested version.
- Checksum has been changed to use CRC32 for plugin syncs
### Checklist before requesting a review


.github/workflows/bench.yml
```diff
@@ -61,9 +61,8 @@ on:
         default: "4cpu-16g-1xT4"
 
 env:
-  CLOUDSMITH_ENTITLEMENT_TOKEN: ${{ secrets.CLOUDSMITH_ENTITLEMENT_TOKEN }}
   HF_TOKEN: ${{ secrets.HF_TOKEN }}
-  CEDANA_URL: ${{ vars.CEDANA_URL }}
+  CEDANA_URL: "https://rahul.cedana.ai/v2"
```
yashanand1910: Let's revert once approved.

yashanand1910 (dismissed):

.github/workflows/publish.yml
```diff
@@ -91,118 +91,246 @@ jobs:
...
-            --name cedana-${{ matrix.arch }} \
-            --tags ${{ matrix.arch }},${{ github.head_ref || github.ref_name }},${{ github.sha }} \
-            --version ${{ github.head_ref || github.ref_name }} \
-            --sync-attempts 5 --error-retry-max 5 --error-retry-backoff 1
-
-      - name: Push install-release.sh to Cloudsmith
-        if: matrix.arch == 'amd64' # only push once as it's architecture independent
+      - name: Upload cedana & install-release.sh
+        env:
+          CEDANA_URL: "https://rahul.cedana.ai/v2"
```
yashanand1910: Revert this to use secrets/vars from CI. Also this should be on top like in other workflows.

yashanand1910 (approved):

PR #713, opened 2026-01-30
feat: services for propagator integrated tests

swarnimarun:
!TESTING!


yashanand1910 (changes requested):
Please understand the existing pattern, and use the `download.yml` to pull the image digests/binaries. The existing setup is the way it is so that these tests can run in other repos, all of which use `download.yml` before reusing the test workflows from this repo.

PR #739, opened 2026-03-02
CED-1838: add ci test for slurm

TusharMohapatra07:


.github/workflows/pr.yml
```diff
@@ -60,6 +60,7 @@ jobs:
       helm_chart_version: main
       controller_version: main
       containerd_runtime_version: cedana
+      slurm_version: fix/slurm-dump-errors
```
nravic: point to main before merging

.github/workflows/push.yml
```diff
@@ -54,6 +54,7 @@ jobs:
       helm_chart_version: main
       controller_version: main
       containerd_runtime_version: cedana
+      slurm_version: fix/slurm-dump-errors
```
nravic: same as ^

.github/workflows/test.yml
```diff
@@ -523,10 +528,25 @@ jobs:
...
+    secrets: inherit
+    with:
+      ref: ${{ inputs.ref }}
+      debug: ${{ inputs.debug }}
+      large: ${{ inputs.large }}
+      runner: ${{ inputs.runner }}
+      runner_gpu: ${{ inputs.runner_gpu }}
+      parallelism: 1
+      post_summary: false # Will post consolidated summary instead
+      slurm_version: "fix/slurm-dump-errors"
```
nravic: same as ^

.github/workflows/test_slurm.yml
```diff
@@ -0,0 +1,327 @@
...
+          fi
+
+      - name: Install slurm binaries
+        run: |
+          chmod +x slurm-bin/build/cedana-slurm
+          sudo cp slurm-bin/build/cedana-slurm /usr/local/bin/cedana-slurm
+          sudo cp slurm-bin/build/libslurm-cedana.so /usr/local/lib/libslurm-cedana.so
+          sudo cp slurm-bin/build/task_cedana.so /usr/local/lib/task_cedana.so
+          sudo cp slurm-bin/build/cli_filter_cedana.so /usr/local/lib/cli_filter_cedana.so
+
```
nravic, line 168: missing latest job_submit plugin

plugins/slurm/internal/job/utils.go
```diff
@@ -4,12 +4,54 @@ import (
...
+				log.Debug().Uint32("pid", pid).Str("path", path).Msg("resolved cgroup path from /proc")
+				return path, nil
+			}
+			log.Warn().Uint32("pid", pid).Str("path", path).Msg("cgroup path from /proc does not exist on filesystem")
+		} else {
+			log.Debug().Uint32("pid", pid).Err(err).Msg("failed to read cgroup path from /proc, falling back to well-known paths")
+		}
+	}
+
+	// Fallback: check well-known cgroup v2 paths
```
nravic, line 54: I think this is legacy? can you recheck why this is being added?

nravic (commented):
Added comments

.github/workflows/download.yml
```diff
@@ -440,3 +445,44 @@ jobs:
...
+        run: |
+          mkdir -p slurm/build
+          cd slurm
+          VERSION=$(echo $VERSION | sed 's/\//-/g')
+          if [ "${VERSION#v}" != "$VERSION" ] || [ "$VERSION" = "latest" ]; then
+            REPO="cedana/cedana-slurm"
+          else
+            REPO="cedana/cedana-slurm-alpha"
+          fi
+          for binary in cedana-slurm libslurm-cedana.so task_cedana.so cli_filter_cedana.so; do
```
yashanand1910: Can you update the list of binaries, I believe `job_submit` plugin is missing? Please check the latest list of binaries produced by `cedana-slurm`.

.github/workflows/test.yml
yashanand1910: Shouldn’t need to specify the plugin version here. Specifying in the download step should be sufficient, just like it’s for other tests.

.github/workflows/test_slurm.yml
```diff
@@ -0,0 +1,333 @@
+name: Test Slurm
```
yashanand1910: For conciseness in appearance. (slurm gets appended anyway)
```suggestion
name: Test
```

.github/workflows/test_slurm.yml
```diff
@@ -0,0 +1,333 @@
...
+        type: number
+        description: "Parallelism for running tests"
+        required: false
+        default: 1
+      post_summary:
+        type: boolean
+        description: "Post test summary"
+        required: false
+        default: false
+      slurm_version:
```
yashanand1910: Why is this here? The download step is sufficient.

.github/workflows/test.yml
```diff
@@ -523,10 +528,25 @@ jobs:
...
+    uses: ./.github/workflows/test_slurm.yml
+    if: ${{ inputs.slurm }}
+    secrets: inherit
+    with:
+      ref: ${{ inputs.ref }}
+      debug: ${{ inputs.debug }}
+      large: ${{ inputs.large }}
+      runner: ${{ inputs.runner }}
+      runner_gpu: ${{ inputs.runner_gpu }}
+      parallelism: 1
```
yashanand1910, line 550: Please add a comment explaining why we are not passing in the parallelism from this job and forcing 1 (even if it’s temporary).

.github/workflows/test_slurm.yml
```diff
@@ -0,0 +1,333 @@
...
+env:
+  CEDANA_URL: ${{ vars.CEDANA_URL }}
+  CEDANA_AUTH_TOKEN: ${{ secrets.CEDANA_AUTH_TOKEN }}
+  CEDANA_LOG_LEVEL: debug
+  CEDANA_PROFILING_ENABLED: true
+  CEDANA_METRICS_ENABLED: true
+  AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
+  AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+  AWS_REGION: ${{ vars.AWS_REGION }}
+  FORMATTER: pretty
```
yashanand1910: Please check other test workflows, we use `junit` so a summary can be generated and posted to PR/Slack when needed.

.github/workflows/test_slurm.yml
```diff
@@ -0,0 +1,333 @@
...
+      - uses: runs-on/action@v1
+
+      - uses: actions/checkout@v4
+        with:
+          repository: cedana/cedana
+          ref: ${{ inputs.ref }}
+          fetch-depth: 0
+          fetch-tags: true
+
+      - name: Checkout cedana-slurm
```
yashanand1910: We should get rid of this, and download the ansible stuff in `download.yml` itself along with the slurm binaries. Thus, we can get rid of the redundant `slurm_version` input var.

.github/workflows/test_slurm.yml
```diff
@@ -0,0 +1,333 @@
...
+
+      - name: Install slurm binaries
+        env:
+          ARCH: ${{ matrix.arch }}
+        run: |
+          for f in slurm-bin/*-$ARCH; do
+            [ -f "$f" ] || continue
+            dest=$(basename "$f" "-$ARCH")
+            if [ "$dest" = "cedana-slurm" ]; then
+              sudo install -m 755 "$f" /usr/local/bin/cedana-slurm
```
yashanand1910: No need for this step? The above `cedana plugin install slurm` installs the binaries in the `/usr/local/lib` and `/usr/local/bin`.

.github/workflows/test_slurm.yml
```diff
@@ -0,0 +1,333 @@
...
+            else
+              sudo install -m 755 "$f" /usr/local/lib/"$dest"
+            fi
+          done
+
+      - name: Install ansible and Docker CLI
```
yashanand1910, line 169: We always avoid installing anything as part of a CI step.

Please make this a part of the `cedana/cedana-test` docker images. Docker CLI is probably already a part of it. Other option is to install it as part of a BATS lifecycle step (like `setup_suite`). This is **important** because that way we can run all the tests locally as they run inside this container.

.github/workflows/test_slurm.yml
```diff
@@ -0,0 +1,333 @@
...
+      - name: Mark git dir as safe
+        run: git config --global --add safe.directory "$(pwd)"
+
+      - name: Setup debugging session
+        uses: mxschmitt/action-tmate@v3
+        if: inputs.debug
+        with:
+          limit-access-to-actor: true
+
+      - name: Setup SLURM cluster
```
yashanand1910, line 182: Why is this part of a CI step? Please check `test_k8s.yml`, nothing extra is done in the workflow except downloading specific binaries and setting the environment variables before simply running the test using `make`. Please ensure slurm tests are runnable locally. Check out how k8s tests run locally (start with the `makefile`), slurm tests should run the same way.

plugins/slurm/internal/job/utils.go
```diff
@@ -4,12 +4,54 @@ import (
...
+	if err != nil {
+		return "", err
+	}
+
+	for _, line := range strings.Split(strings.TrimSpace(string(data)), "\n") {
+		parts := strings.SplitN(line, ":", 3)
+		if len(parts) != 3 {
+			continue
+		}
+		if parts[0] == "0" && parts[1] == "" {
```
yashanand1910, line 28: So we don’t support cgroup v1? For cgroup v1 `parts[1] == “”` will never be true?

plugins/slurm/internal/job/utils.go
yashanand1910: This is the old logic just being used as fallback so I think it’s ok.

Makefile
```diff
@@ -269,6 +269,21 @@ test-k8s: ## Run kubernetes e2e tests (PARALLELISM=<n>, GPU=[0|1], TAGS=<tags>,
 		fi ;\
 	fi
 
+test-slurm: ## Run slurm e2e tests (PARALLELISM=<n>, GPU=[0|1], TAGS=<tags>, RETRIES=<retries>, DEBUG=[0|1])
```
yashanand1910, line 272: Should automatically run inside a container just like `test-k8s`, so that when run locally no dependency is needed. And on CI, it runs directly instead of inside the container. Please check `make test-k8s`.

.github/workflows/test_slurm.yml
yashanand1910: You can update the test image and run the `Manual (publish)` workflow to update the official `cedana/cedana-test` images. But don’t do this until you’re able to test locally. The whole test suite should be able to run locally.

yashanand1910 (changes requested):

nravic (dismissed):
fantastic work! just change the target branch and good to merge.

nravic (approved):

PR #789, opened 2026-05-14
CED-2073: Use New CRIU hook for regular processes

bsach64:
## Describe your changes
Requires
- https://github.com/cedana/criu/pull/21
- https://github.com/cedana/cedana-gpu/pull/233

We add a new hook to CRIU for regular processes at the point it which it skips namespace restore (CRIU pr does this). This allows us to  attach gpu controller earlier than PostRestore for regular processes
For Containers, we now dump env vars as part of the gpu checkpoint, allowing us to attach at SetupNamespaces once again.

### DIFF
--- a/internal/cedana/gpu/controller.go
+++ b/internal/cedana/gpu/controller.go
@@ -569,7 +569,7 @@ func (p *pool) CRIUCallback(id string) *criu_client.NotifyCallback {
 
 	// Update GPU controller with the restore PID (which can be a new PID)
 	var restoredPid *int32
-	callback.PostRestoreFunc = func(ctx context.Context, pid int32) error {
+	callback.NoNsFunc = func(ctx context.Context, pid int32) error {
 		controller := p.Get(id)
 		restoredPid = &pid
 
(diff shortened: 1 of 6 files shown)


bsach64:
This Greptile comment is incorrect

> CRIU passes the PID of the namespace-init helper process to the setup-namespaces hook, not the PID of the actual root task being restored. Using that PID to call controller.Attach could bind the GPU controller to the wrong process when namespace restore is active. The previous PostRestoreFunc received the root-task PID (as seen from the host namespace), which is the correct attachment target. If GPU workloads can be checkpointed with namespace isolation, restoring them with this code would attach to the init helper and leave the actual GPU process unattached — causing the Attach or later syscall.Kill to operate on the wrong PID.

the same pid is returned regardless of `ACT_SETUP_NS` or `ACT_SKIP_NS` which is that of root task
look at corresponding CRIU code for more details: https://github.com/cedana/criu/pull/21
specifically this code on the pr branch: https://github.com/cedana/criu/blob/feature/ced-2073-start-gpu-restore-earlier-by-storing-env-variables-as-part/criu/cr-service.c#L211-L221

yashanand1910 (approved):

PR #794, opened 2026-05-21
ced-2055 fix path resolution for nvidia-persistenced/socket on restore

LGra78:
## Describe your changes
Fixes runc GPU restore failure when the checkpoint includes NVIDIA persistenced socket. Mount.Root was being resolved as /nvidia-persistenced/socket which is not a valid host path, causing an error. This fix instead used MountPoint when possible restoring the socket from the correct directory /run/nvidia-persistenced/socket. 

- [Y] Have I performed a self-review?
- [NA] Have I added regression or unit tests (where it makes sense) for my changes?
- [NA] Have I updated the docs if changes have resulted in it being out of date?
- [Y] Have I checked to ensure my PR only introduces the changes it's intended to? E.g. No extraneous formatting changes.
- [NA] Have I ensured that there are no performance regressions by checking benchmark results?
- [N] Is this a breaking change? If yes, please describe areas affected and request reviews from developer stakeholders.

### DIFF
--- a/internal/cedana/gpu/restore_adapters.go
+++ b/internal/cedana/gpu/restore_adapters.go
@@ -328,7 +328,7 @@ func AddMountsForRestore(next types.Restore) types.Restore {
 				if resolvedSource, ok := resolvedMountSources[m.ID]; ok {
 					source = resolvedSource
 				}
-				log.Debug().Str("root", m.Root).Str("mount_path", m.MountPoint).Msg("marking NVIDIA GPU mount as external")
+				log.Debug().Str("root", source).Str("mount_path", m.MountPoint).Msg("marking NVIDIA GPU mount as external")
 				req.Criu.External = append(req.Criu.External, fmt.Sprintf("mnt[%s]:%s", m.MountPoint, source))
 			}
 			return true
--- a/internal/cedana/gpu/restore_adapters.go
+++ b/internal/cedana/gpu/restore_adapters.go
@@ -5,6 +5,7 @@ import (
 	"encoding/binary"
 	"fmt"
 	"os"
+	"path/filepath"
 	"regexp"
 	"strconv"
 	"strings"
@@ -18,6 +19,7 @@ import (
 	"github.com/cedana/cedana/pkg/types"
 	"github.com/cedana/cedana/pkg/utils"
 	"github.com/google/uuid"
+	"github.com/moby/sys/mountinfo"
 	"github.com/rs/zerolog/log"
 	"github.com/spf13/afero"
 	"google.golang.org/grpc/codes"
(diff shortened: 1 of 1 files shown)


internal/cedana/gpu/restore_adapters.go
```diff
@@ -330,6 +331,19 @@ func AddMountsForRestore(next types.Restore) types.Restore {
 	}
 }
 
+// nvidiaMountRestoreSource returns the source path CRIU should use for an NVIDIA mount.
+// Prefer MountPoint when it exists; Mount.Root may not be a valid host path.
+func nvidiaMountRestoreSource(m *daemon.Mount) string {
```
yashanand1910: How do we know this is the right solution? Why can `Mount.Root` not be a valid path? Isn't `Mount.Root` supposed to be the path on the host always?

yashanand1910 (commented):
Please check my comments.

internal/cedana/gpu/restore_adapters.go
LGra78: From my testing with nvidia-persistenced, Mount.Root resolved to /nvidia-persistenced/socket, which was not a valid path on the host. The actual socket was located at /run/nvidia-persistenced/socket, which was correctly referenced by MountPoint. Since CRIU needs a valid host path for open_tree during restore, preferring MountPoint when it exists avoids passing an invalid restore source.

yashanand1910: That's fine. I understand that this would fix the issue for `nvidia-persistenced` right now, but does not sufficiently address the issue. `Mount.Root` is always supposed to be right path, and has been so for all other mounts that NVIDIA adds and it has been working for us.

Why is it that only this particular path is incorrect? You mention `Mount.Root` may not be a valid path on host, why is that and when does that happen? If we find an answer to this, I believe we should have a better solution. Right now we are just silently using `Mount.MountPoint` when the root path does not exist and _hoping_ it would work. 

Please check CRIU documentation on this option. (https://criu.org/External_bind_mounts)

> On dump, KEY is a mountpoint inside container, and corresponding VAL is a string that will be written into the image as mountpoint's root value.
> On restore, KEY is the value from the image (VAL from dump), and the VAL is the path on host that will be bind-mounted into container (to the mountpoint path from image). 

According to this, the current logic should work because on restore we need to use `Mount.Root`. It might be worthwhile tracing back to where this `Mount.Root` gets populated and if there's an issue with that.

LGra78: I see what you're saying, yes I agree that could be a better fix, I'll start looking into it. Thanks.

LGra78: Hi @yashanand1910, I've made some changes I'd like your thoughts on.

I traced Mount.Root back to the imported mountinfo.go. Root is defined as the path relative to the mounted filesystem, so since this tmpfs is mounted at /run, it makes sense, and is correct/intentional, that /run is not in the path stored in Mount.Root. MountPoint is relative to the process root, so it includes /run, which is why my earlier fallback to MountPoint worked.

Instead of falling back to MountPoint, I updated the code to match the filesystem using the mount's Major and Minor IDs from mountinfo. Those identify the filesystem instance, so in this case it matches the dumped /nvidia-persistenced/socket mount to the same tmpfs instance mounted at /run, then resolves the host-visible source path CRIU needs: /run/nvidia-persistenced/socket. It is resolved by joining the path string of m.Root with the root of the matched filesystem (in this case, /run + /nvidia-persistenced/socket => /run/nvidia-persistenced/socket).

Current behavior is if m.Root exists, use it, if not, match an external filesystem using the Major and Minor IDs. If no match is found, use m.Root (which will result in a CRIU error, a warning/log may be useful here).

What do you think of this compared to my prior change? Both fall back to using other information in mountinfo to resolve the path, as m.Root intentionally does not contain /run. The difference is which pieces of information in mountinfo are used (prior used m.MountPoint, current uses Major and Minor IDs to match a filesystem mount).

internal/cedana/gpu/restore_adapters.go
```diff
@@ -330,6 +333,34 @@ func AddMountsForRestore(next types.Restore) types.Restore {
 	}
 }
 
+// resolveExternalMountSource returns the host path CRIU should bind mount from.
+// It keeps m.Root when it is already valid, otherwise resolves it through a matching host filesystem mount.
+func resolveExternalMountSource(m *daemon.Mount) string {
```
yashanand1910: To avoid calling this helper function several times, I suggest an optimization. Resolve the correct `m.Root` at the place where we populate all mounts. https://github.com/cedana/cedana/blob/9768a66055f3659159335f11cedfa0e30be9973d/pkg/utils/process.go#L193-L213
This can be done by looping through all the mounts only once instead of for each mount.

yashanand1910 (changes requested):

internal/cedana/gpu/restore_adapters.go
LGra78: @yashanand1910 Made the optimization by building a map of resolved sources once, so the mount list is only looped through once up front instead of once per NVIDIA mount. The map is then read by m.ID when adding the CRIU external mounts.

I kept this separate from the m.Root logic instead of overwriting it because I think m.Root is already correct as is. Per the earlier comment, based on mountinfo.go, it intentionally represents path relative to mounted filesystem, so it should not include the mounted filesystem location (i.e. /run).

What do you think?

yashanand1910: The optimization's fine. Can we also uplift this logic to where the `Mount` struct is populated. Check the the link I sent in my earlier comment. That way, all users of this struct have the correct root.

LGra78: @yashanand1910 Sure, to confirm: should Mount.Root be populated as the host-visible path, including the filesystem mount location like /run, so consumers can treat it as a full host path rather than the current mountinfo definition of root relative to the mounted filesystem?

yashanand1910: Yes, it should always be the correct path instead of the relative one, using this same logic here.

LGra78: @yashanand1910  I tried moving this to where Mount.Root is populated, but globally changing m.Root revealed other restore errors, for example: "Failed to open_tree /sys/fs/cgroup/system.slice/runc-nvidia-smi4: No such file or directory".

I think this indicates other parts of the restore path depend on the current mountinfo definition of m.Root. In this case, cgroup restore expects the cgroup root relative to the cgroup filesystem, not rewritten as a host-visible /sys/fs/cgroup/... path. This was the first issue I hit after overwriting m.Root, so I’m hesitant to assume cgroup is the only case affected.

Because of that, I think it makes more sense to keep this fix limited to restore_adapters.go, where we specifically need a host visible source path for CRIU external bind mount restore. What do you think?

PR #802, opened 2026-05-29
chore(slurm): handle rmq state logging

swarnimarun:


nravic (approved):

yashanand1910:
All returned errors are always captured and logged when they are called by `cedana-slurm`.

PR #814, opened 2026-06-10
CED-2138: add preemption c/r sanity test and remove gpu tests in the slurm ci

TusharMohapatra07:
### Preemption test
  - Add `test/slurm/preemption.bats` — submit a victim job, preempt it, verify checkpoint + restore.
  - Add a dedicated `Slurm (Preemption)` CI job (`test_slurm.yml`); tag it out of the main SLURM run (`!preemption`).
  
  ### SLURM monitor / cluster-id
  - CI (`slurm_setup.bash`) persists the cluster ID into `/etc/cedana/config.json` on each node via `cedana --merge-config` after registration.
  
  ### SLURM cgroup path
  - Resolve the job cgroup from `/proc/<pid>/cgroup` (v2/v1 fallbacks) instead of hostname-based path guessing, which breaks under nested `--cgroupns=host`.
  
  ### GPU SLURM tests removed
  - Delete `test/slurm/gpu.bats`; drop `CUDA` from the SLURM test matrix.
  
target cedana-slurm branch: https://github.com/cedana/cedana-slurm/pull/136


yashanand1910 (changes requested):
Why are we rebuilding the binaries here? The process to follow is to simply change your pr.yml with the cedana_version or slurm_version depending on which repo you're working on if you have a PR that is on both repos. Download step must just download pre-built binaries from cloudsmith. Otherwise, we will end up hiding issues with version skew.

.github/workflows/test_slurm.yml
```diff
@@ -706,6 +704,239 @@ jobs:
...
+
+      - name: Extract cedana-slurm
+        run: mv slurm-bin/cedana-slurm cedana-slurm
+
+      - name: Mark git dir as safe
+        run: git config --global --add safe.directory "$(pwd)"
+
+      - name: Pre-install plugins
+        env:
+          CEDANA_PLUGINS_BUILDS: alpha
```
nravic: ?

PR #816, opened 2026-06-10
Add GPU dump and restore profiling display

LGra78:
## Describe your changes
- Read `GpuProfile` from GPU dump responses.
- Read `GpuProfile` from GPU restore responses.
- Add GPU function profiling rows into Cedana profiling output.
- Display function rows under the normal `gpu` profiling category.
- Add worker profiling rows:
  - `workerSummary`
  - `workerSlowest`
  - `worker`
  - `workerPhaseAverage`
  - `workerPhaseSlowest`
- Display per-worker bytes without double-counting those bytes in the profiling footer.
- Mark display-only GPU byte rows so they do not inflate profiling totals.
- Keep parent profiling rows visible next to their child rows.
- [NA] Have I added regression or unit tests (where it makes sense) for my changes?
- [NA] Have I updated the docs if changes have resulted in it being out of date?
- [NA] Have I ensured that there are no performance regressions by checking benchmark results?

Not a breaking change. New functionality requires updated gpu plugin, but old functionality will work without it


internal/cedana/gpu/controller.go
```diff
@@ -68,6 +68,103 @@ const (
 	MAX_SYNC_FAILURES = 2
 )
 
+func addGPUFunctionProfileToProfiling(ctx context.Context, duration time.Duration, parallel bool, f ...any) context.Context {
```
yashanand1910: We should move all these helpers to a `profiling.go` within the `gpu` package. `controller.go` is getting quite large.

pkg/profiling/data.go
```diff
@@ -23,25 +23,27 @@ type Data struct {
 	IO       int64 `json:"io,omitempty"`
 	// add more othogonal fields here as needed
 
-	Parallel  bool `json:"parallel,omitempty"`
-	Redundant bool `json:"redundant,omitempty"`
+	Parallel    bool `json:"parallel,omitempty"`
+	Redundant   bool `json:"redundant,omitempty"`
+	IORedundant bool `json:"io_redundant,omitempty"`
```
yashanand1910, line 29: What's this change for?

yashanand1910 (changes requested):

pkg/profiling/data.go
LGra78: If marked IO redundant it does not get added to the running IO total (read/write) - this is helpful for visualizing bytes read/wrote for some rows that were requested

internal/cedana/gpu/controller.go
LGra78: This is a good idea, thanks!

nravic (approved):
lgtm, great work!

PR #817, opened 2026-06-11
Feature/ced 2154 grpc message size blow up

Finn-Andersonn:
## Describe your changes
The info from scraping /proc for the Mistral 123B model on the 8xRTXPro6000 node was too big for the default gRPC recvmsg size. On top of that, the size by which it exceeded would grow consistently with every subsequent failed call.

Just updated mountInfo to match how we append it to a longstanding ProcessState, instead of a constantly increasing list. Then upped the recvmsg size to 6MiB.


yashanand1910 (changes requested):
Please check greptile's comments.

PR #820, opened 2026-06-18
CED-2118: parallelize restore for containers

bsach64:
## Describe your changes
This PR only has the glue code necessary to send the CRIU PostSetupNamespace response to the gpu-controller Attach Request. This allows gpu controller to use `criu_pid, mnt_ns_fd, root_fd` to mount checkpoint dir into the mount namespace of the process being restored (using `mnt_ns_fd`) and also spawn workers in the specified `mnt_ns_fd`.

**EDIT**: We don't error out in case PostSetupNamespaces is called without mount info because a CRIU restore **always** creates a time namespace and means any subsequent checkpoint will dump a time namespace, hence the Post Setup Namespace hook can be triggered for non-containers workloads as well (look at this for more details: https://linear.app/cedana/issue/CED-2118/parallelize-restore-for-containers#cr-cr-edge-case-with-d0dc637a)

Requires:
- https://github.com/cedana/cedana-api/pull/84
- https://github.com/cedana/criu/pull/28
- https://github.com/cedana/cedana-gpu/pull/245

Testing:
- Verified that `vector_add`, `vector_add_host`, `mem-throughput-saxpy-loop`, `vector_add_multi_gpu` all pass when run using the `test/workloads/bundle_cuda/config.json` with the runc plugin
- manually recreated the race that previously occurred when using PostSetupNamespaces and verified that it does not occur with these new changes. (look at ticket for more details)
- verified that the mount namespace fds have the same inode number in both CRIU and GPU (look at ticket for more details)
- verified all restore,runc tests pass
```
make test-regression TAGS="\!large,restore,runc" PARALLELISM=1 GPU=1 DEBUG=1
```

- [Not Required] Have I added regression or unit tests (where it makes sense) for my changes?
- [Not Required] Have I updated the docs if changes have resulted in it being out of date?


internal/cedana/gpu/controller.go
```diff
@@ -562,15 +562,21 @@ func (p *pool) CRIUCallback(id string) *criu_client.NotifyCallback {
...
+		mountNsInfo := &MountNsInfo{}
+		if postSetupNsResp != nil {
+			mountNsInfo.CriuPid = uint32(postSetupNsResp.GetCriuPid())
+			mountNsInfo.MntNsFd = uint32(postSetupNsResp.GetMntNsFd())
+			mountNsInfo.RootFd = uint32(postSetupNsResp.GetRootPathFd())
+			log.Debug().Any("mountNsInfo", mountNsInfo).Msg("got mountNsInfo of process from CRIU")
+			return controller.Attach(ctx, uint32(pid), mountNsInfo)
+		}
+		return fmt.Errorf("could not get mount namespace info from CRIU")
 	}
```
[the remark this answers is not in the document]
bsach64: This is completely fine and expected.

internal/cedana/gpu/controller.go
```diff
@@ -644,6 +650,15 @@ func (p *pool) Check(binary string) types.Check {
 /// CONTROLLER ///
 //////////////////
 
+// specifies information gpu-controller uses
+// to enter in the mount namespace of the
+// process being restored to mount checkpoint dir.
+type MountNsInfo struct {
```
yashanand1910: Redundant type. Please remove this and use the proto type directly.

internal/cedana/gpu/controller.go
```diff
@@ -722,8 +737,16 @@ func (c *controller) Sync(ctx context.Context, wait bool) (err error) {
 }
 
 // Forcefully attach to a PID, so that on next Info call, the controller will return this as the attached PID.
-func (c *controller) Attach(ctx context.Context, pid uint32) (err error) {
-	_, err = c.ControllerClient.Attach(ctx, &gpu.AttachReq{PID: pid})
+func (c *controller) Attach(ctx context.Context, pid uint32, mountNsInfo *MountNsInfo) (err error) {
```
yashanand1910: Can make `mountNsInfo` an optional argument to keep it's usages cleaner at other locations.

yashanand1910 (changes requested):

internal/cedana/gpu/controller.go
bsach64: done

internal/cedana/gpu/controller.go
bsach64: Go doesn't have optional arguments, i made a separate `AttachWithMountNs` function to keep usage cleaner at other locations
i don't completely understand what you mean here

PR #835, opened 2026-07-15
CED-2236: Add unprivileged SLURM c/r in CI

TusharMohapatra07:
### .github/workflows/test_slurm.yml
New Slurm (Unprivileged) job with a root_squash / no_root_squash matrix.

### internal/cedana/criu/generic_adapters.go
Reorder criu path resolution so a configured CRIU binary path (config.CRIU.BinaryPath) takes precedence over the installed plugin path.

### internal/cedana/criu/checks.go, checks_cuda.go
Same precedence reorder in the CRIU health/feature checks so the check runs against the same binary the dump will use.

### test/helpers/slurm_setup.bash
- setup_slurm_unprivileged_user + restart_cedana_slurm_daemon_unprivileged: create the submit user, stage node-local capped cedana/criu, and bake plugins.bin_dir / criu.binary_path into /etc/cedana/config.json via --merge-config.
- chmod 666 the monitor log so the uid-2000 monitor's marker lands for the assert.
-  start_cedana_slurm_daemon: DB-readiness gate (MariaDB ping + slurmdbd.conf + sacctmgr) before launching the daemon, so SyncJobs isn't permanently disabled by a not-yet-ready accounting DB.

### test/helpers/slurm_propagator.bash
- checkpoint_slurm_job now sends cluster_id in the request (required for the propagator's cluster-scoped job lookup).

### test/slurm/cpu.bats
Minor tag cleanup on the samples (password-hashing runs unprivileged now that the sample no longer needs root).


nravic:
Add ticket name + update description please. This looks good though! Should also add it to the preemption test.

nravic (dismissed):
Remove placeholders.

PR #842, opened 2026-07-21
feat(s3): support EKS Pod Identity credentials

swarnimarun:
- Pod Identity host-path handling
  - systemd credential environment propagation
  - Support for PodIdentity on EKS

related: https://github.com/cedana/cedana-helm-charts/pull/124
shim version bump (will need an update after this is merged): https://github.com/cedana/cedana-containerd-runtime/pull/67

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -318,6 +318,12 @@ func setDefaults() {
 	for _, field := range utils.ListLeaves(Config{}) {
 		tag := utils.GetTag(Config{}, field, FILE_TYPE)
 		defaultVal := utils.GetValue(Global, field)
+		// Keep the static credential mode as an in-memory default, but don't
+		// persist the newly added key unless it was explicitly configured.
+		// Older runtime shims use exact config decoding and reject unknown keys.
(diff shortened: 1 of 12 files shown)


nravic (dismissed):

pkg/config/config.go
```diff
@@ -313,6 +318,12 @@ func setDefaults() {
 	for _, field := range utils.ListLeaves(Config{}) {
 		tag := utils.GetTag(Config{}, field, FILE_TYPE)
 		defaultVal := utils.GetValue(Global, field)
+		// Keep the static credential mode as an in-memory default, but don't
+		// persist the newly added key unless it was explicitly configured.
+		// Older runtime shims use exact config decoding and reject unknown keys.
+		if tag == "aws.credentials_mode" && defaultVal == DEFAULT_AWS_CREDENTIALS_MODE {
```
yashanand1910, line 324: Don't understand this change. Is this just for backward compatibility with older shims?

yashanand1910 (changes requested):

pkg/config/config.go
swarnimarun: Yeah.

yashanand1910 (dismissed):
Approved conditionally. Let's removed the backwards compat bit in config.go.

PR #843, opened 2026-07-27
ced-2149 add theoretical limit profiling and measurements

LGra78:
## Describe your changes
- `cedana measure` runs storage and memory copy benchmarks. On NUMA hosts it also runs the full local and remote NUMA memory copy benchmarks. It uses `/tmp/cedana-measure` for storage tests unless `--benchmark-path` is set provided.
- `--benchmark-size-gb` sets the amount of data used for each storage and memory test (default: 1)
- `--benchmark-samples` sets how many runs are taken for each test (default 3). The median result is reported.
- `numactl` is optional. Without it, storage and normal host memory results stil run, and NUMA portion is reported unavailable.
- Benchmarks are intentionally under `cedana measure` rather than regular checkpoint/restore profiling. Running them with every checkpoint would add significant noise.
- Theoretical limits are part of regular checkpoint/restore profiling and come from cedana-gpu.

Benchmarks in cedana measure:
- read_cached: reads test file immediately after writing it. Data is expected to be in the Linux page cache, so this mostly shows buffered read path rather than physical disk throughput
- write_cached: writes through normal buffered file I/O and does not call fsync. Measures how quickly data is accepted into the page cache, not when it reaches durable storage.
- read_cold: asks Linux to drop this file's cached pages with posix_fadvise(..., FADV_DONTNEED) before reading. This is closer to storage read throughput, although the OS may not be guaranteed to evict the cached pages.
- write_durable: performs the same write as write_cached but calls fsync before stopping the timer. This includes waiting for the filesystem to make the write durable.
- host_memory_copy_default: copies one allocated buffer to another using the default OS memory placement policy.
- host_memory_copy_local: runs the copy with CPU and memory bound to the same NUMA node.
- host_memory_copy_remote: runs the CPU bound to one NUMA node and memory allocated to another, e.g. numa0->numa1.

Example theoretical profiling output:
`NVIDIA A100-SXM4-80GB (GPU-599a4549-c008-947d-5b30-e913b8a0dcb1) limits: device memory 2039 GB/s, host link 31.5 GB/s`

Example Cedana measure commands:
`cedana measure` (defaults)
`cedana measure --benchmark-path /mnt/checkpoints --benchmark-size-gb 2 --benchmark-samples 5` (specifying benchmark size, sample count, and path to use for storage benchmarks)

Example cedana measure output:

```bash
SECTION  NAME          METRIC                    GB/S  CAPACITY GB  TEST GB  SAMPLES  
storage  /             read_cached                5.4        659.2        1        3  
storage  /             write_cached               2.5        659.2        1        3  
storage  /             read_cold                  1.9        659.2        1        3  
storage  /             write_durable              0.8        659.2        1        3  
memory   shadecloud    host_memory_copy_default  11.1        105.6        1        3  
memory   numa0         host_memory_copy_local    10.4         52.8        1        3  
memory   numa0->numa1  host_memory_copy_remote   11.1         52.8        1        3
memory   numa1->numa0  host_memory_copy_remote    8.6         52.8        1        3
memory   numa1         host_memory_copy_local    11.1         52.8        1        3
```


cmd/measurements.go
```diff
@@ -0,0 +1,282 @@
...
+	measurementsCmd.Flags().Bool("collect-numa", false, "collect NUMA topology rows")
+	measurementsCmd.Flags().Bool("benchmark-numa", false, "benchmark NUMA-local memory copy paths")
+	measurementsCmd.Flags().Bool("full-numa-matrix", false, "benchmark all NUMA source/target memory paths")
+
+	benchmarkMeasurementsCmd.Flags().String("result-file", "", "write the memory rate to a file")
+	benchmarkMeasurementsCmd.Flags().Float64("size-gb", 1, "benchmark working set size in GB")
+}
+
+var measurementsCmd = &cobra.Command{
+	Use:   "measurements",
```
yashanand1910: Let's just call it "measure".

pkg/measurements/measurements.go
```diff
@@ -0,0 +1,330 @@
...
+	Name            string
+	Backend         string
+	CapacityGB      float64
+	Mode            string
+	CachePolicy     string
+	BenchmarkSizeGB float64
+	BenchmarkRuns   int
+	ReadGBPerSec    *float64
+	ReadSource      string
+	ReadConfidence  string
```
yashanand1910: What's this?

pkg/measurements/measurements.go
```diff
@@ -0,0 +1,330 @@
...
+	"slices"
+	"strings"
+	"syscall"
+
+	"buf.build/gen/go/cedana/cedana/protocolbuffers/go/daemon"
+	"github.com/cedana/cedana/pkg/utils"
+	"github.com/shirou/gopsutil/v4/disk"
+)
+
+const (
```
yashanand1910, line 18: Let's instead have a simple `proto` type, so that the GPU-specific measurements, or any other measurements that don't belong here can be easily sent to the daemon without making too many modifications to the type in the future.

yashanand1910 (changes requested):
1. Simplify types and define new `proto` types in `cedana-api`.
2. GPU measurements belong in `cedana-gpu` repository.

For reference, just look at how we exchange health check data type: https://github.com/cedana/cedana-api/blob/main/cedana/daemon/health.proto

cmd/measurements.go
yashanand1910, line 1: File should be called `measure.go`. Please check other subcommand files.

cmd/measurements.go
```diff
@@ -0,0 +1,221 @@
...
+	"github.com/cedana/cedana/pkg/style"
+	"github.com/jedib0t/go-pretty/v6/table"
+	"github.com/jedib0t/go-pretty/v6/text"
+	"github.com/spf13/cobra"
+)
+
+func init() {
+	measureCmd.AddCommand(benchmarkMeasureCmd)
+
+	measureCmd.Flags().String("benchmark-path", measurements.DefaultBenchmarkPath, "storage path to benchmark")
```
yashanand1910, line 19: Why are we not using `pkg/flags`?

cmd/measurements.go
```diff
@@ -0,0 +1,221 @@
...
+	measureCmd.Flags().Float64("benchmark-size-gb", 1, "benchmark working set size in GB")
```
yashanand1910, line 20: Why are the flags so verbose? Just `--size-gb` is much simpler. It's not going to conflict with anything. Same for samples and path.

cmd/measurements.go
```diff
@@ -0,0 +1,221 @@
...
+	Use:   "measure",
+	Short: "Collect and print node throughput measurements",
+	Args:  cobra.NoArgs,
+	RunE: func(cmd *cobra.Command, args []string) error {
+		report := &measurements.Report{}
+		benchmarkPath, _ := cmd.Flags().GetString("benchmark-path")
+		benchmarkSizeGB, _ := cmd.Flags().GetFloat64("benchmark-size-gb")
+		benchmarkSamples, _ := cmd.Flags().GetInt("benchmark-samples")
+
+		storage, err := measurements.BenchmarkStorage(cmd.Context(), benchmarkPath, benchmarkSizeGB, benchmarkSamples)
```
yashanand1910, line 38: Let's make all these subcommands, so one can call `cedana measure storage ...`. Then you can simply add a `cedana measure all ...` that does all.

cmd/measurements.go
```diff
@@ -0,0 +1,221 @@
...
+		if err != nil {
+			return err
+		}
+		report.NUMA = numa
+		printMeasurementsTable(report)
+		return nil
+	},
+}
+
+var benchmarkMeasureCmd = &cobra.Command{
```
yashanand1910, line 58: Can you add an example in the description for an output of this?

yashanand1910 (changes requested):
Left some comments. Also, please add more examples in the description.

Sorry, but this probably needs to be redesigned. The design is confusing. Theoretical limits needs to be a more generic part of Cedana, and not just used for the GPU plugin. Let's set up a sync to rethink this design.

PR #846, opened 2026-08-03
fix(CED-2260): hard copy bins instead of failing

swarnimarun:
Today it fails copy if the cedana binary is already in use, and stop didn't work properly I think it's better to just ignore the os and hard copy.


rahulk789 (dismissed):

yashanand1910 (dismissed):
This is great @swarnimarun. Can you also fix other scripts which have the same issue: https://github.com/cedana/cedana/blob/a10282621120a28ef6728f9219d318a807e7eb42/scripts/install-release.sh#L65-L69

yashanand1910 (approved):

PR #847, opened 2026-08-03
CED-2200: Directory support for CSX

bsach64:
## Describe your changes

Extends `io.Storage` to have a couple for methods to allow plugins to provide paths on the filesystem based on the checkpoint name which can be passed to GPU controller + CRIU, so that we c/r from a path provided by the plugin.

`CreatePath()` which takes in a `Dir` and a `Name` and returns a corresponding path and a cleanup() function which should be called when the client finishes writing to the provided path.

In case of CSX `Dir` is `csx://` and `Name` would be a UUID (k8s checkpoint) and CSX would return something like
`/dir/UUID`

`ReadPath()` takes in a "path" which would something like `csx://UUID` and CSX would return `/cache/UUID` from which the client can read the checkpoint.

`cleanup()` allows for the plugin to run a function after client is done reading/writing from a path, so in case of CSX it allows the CSX plugin to call `ClosePath()` and inform CSX that the read/write is done.

In case of compression and/or if the plugin's `storage.Remote() = true` we always provide `os.TempDir` to CRIU + gpu controller and we only provide the final compressed artifact to the plugin i.e CSX by using `Storage.Create()` 

Api which the csx plugin calls: https://github.com/cedana/cedana-api/pull/87


yashanand1910 (dismissed):
Approved provisionally. Storage interface to be simplified/unified later.

PR #849, opened 2026-08-06
CED-2132: refresh CRIU external files after freeze

LGra78:
## Describe your changes
Fixes CED-2132, where a process dump could fail if a process kept an FD open
while its mount disappeared after Cedana read `/proc`, but before CRIU froze
the process tree. Cedana previously built CRIU's external file list from that
initial state, so the disappeared mount was missing when CRIU needed it. This
keeps that initial list for the existing path, then rebuilds it when CRIU sends
`query-ext-files` after freezing the tree, so CRIU gets the current list
without changing the normal behavior.

- [x] Have I added regression or unit tests (where it makes sense) for my changes? N/A existing automated tests caught this bug
- [x] Have I updated the docs if changes have resulted in it being out of date? N/A internal CRIU fix
- [x] Have I ensured that there are no performance regressions by checking benchmark results? NA
- [x] Is this a breaking change? If yes, please describe areas affected and request reviews from developer stakeholders. Non breaking change

### DIFF
--- a/internal/cedana/gpu/dump_adapters.go
+++ b/internal/cedana/gpu/dump_adapters.go
@@ -6,6 +6,7 @@ import (
 	"regexp"
 
 	"buf.build/gen/go/cedana/cedana/protocolbuffers/go/daemon"
+	"github.com/cedana/cedana/pkg/criu"
 	"github.com/cedana/cedana/pkg/types"
 	"github.com/cedana/cedana/pkg/utils"
 	"github.com/rs/zerolog/log"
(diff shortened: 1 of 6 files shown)


yashanand1910 (changes requested):
Can we check for all the remaining cases where `req.Criu.External` is updated? You will likely have to do a usage search on this field to catch all cases. For e.g.

https://github.com/cedana/cedana/blob/16e65dff4ce8558b1a6a04a84b708bd53f5a7dc2/internal/cedana/gpu/dump_adapters.go#L81-L87

https://github.com/cedana/cedana-runc/blob/81742d8b409a0ef84f2241d1c7d514a35c702e8b/internal/device/dump_adapters.go#L32-L36

etc.

LGra78:
> Can we check for all the remaining cases where `req.Criu.External` is updated? You will likely have to do a usage search on this field to catch all cases. For e.g.
> 
> https://github.com/cedana/cedana/blob/16e65dff4ce8558b1a6a04a84b708bd53f5a7dc2/internal/cedana/gpu/dump_adapters.go#L81-L87
> 
> https://github.com/cedana/cedana-runc/blob/81742d8b409a0ef84f2241d1c7d514a35c702e8b/internal/device/dump_adapters.go#L32-L36
> 
> etc.

I did a pass over the remaining `req.Criu.External` writes. GPU mounts were the only other entries derived from the pre-freeze `ProcessState`, so I added a `query-ext-files` callback there as well.

The other dump-time uses are built from fixed CRIU options or container configuration rather than target process state, so they are not affected by this stale state window, so I did not change them.

@yashanand1910 what do you think?

yashanand1910 (approved):

PR #851, opened 2026-08-10
CED-2258: Mount nvme into bench container

nravic:
## Describe your changes
OOM/OOS/OOH!


yashanand1910 (changes requested):
That path is not guaranteed to exist for every spawnable runner. What you likely need to do is add `preinstall` commands to the runner definitions in `runs-on.yml` in our `.github-private` repo.

You can add them for all the `T4` node definitions now. The definition should basically setup docker to use the attached volume that comes with the node. Or, maybe even simply increasing the volume size might do it.

PR #854, opened 2026-08-14
feat(k8s): provide /checkpoint as tmpfs for the memory storage tier

swarnimarun:
## What

Provides `/checkpoint` as tmpfs for the memory storage tier, via a
`configure-checkpoint-tmpfs.sh` helper script wired into the k8s plugin.

## Why

The memory storage tier has to be memory-backed to mean anything. This is what
makes a 42 GiB CRIU + GPU checkpoint complete in about 30 seconds with the model
still serving; on a disk-backed path that figure does not hold, and the checkpoint
competes with the very workload it is trying to snapshot.

It also constrains restore: a memory-tier checkpoint is node- and boot-local,
which is why the controller's restore gate checks the node's boot id before
trusting one.

## Stacking

Base is `swarnim/k8s-helper-image-plugins` (PR 2 of 2 in this repo's stack).

## Depends on

- **cedana-helm-charts `swarnim/nebius-checkpoint-tmpfs`** — configures this on the
  Nebius GPU nodes.
- **cedana-controller `swarnim/dynamo-checkpoint-restore`** — the consumer.

## Note

`plugins/runc` is a submodule in this repo and its pointer is **not** advanced by
this PR. The `/dev/shm` archive/restore work that the restore path needs lives in
`cedana/cedana-runc` and is still uncommitted locally — it needs its own PR there
first, after which the submodule pointer can be bumped in a separate change.

### DIFF
--- a/plugins/k8s/cmd/helper.go
+++ b/plugins/k8s/cmd/helper.go
@@ -76,6 +76,11 @@ var setupCmd = &cobra.Command{
 			script.Chroot("/host", k8scripts.ConfigureKubelet),
 			script.Chroot("/host", scripts.ConfigureShm),
 			script.Chroot("/host", scripts.ConfigureIoUring),
+			// Runs before the service so /checkpoint exists before anything can
+			// be asked to write a checkpoint into it. Disabled unless
+			// CHECKPOINT_TMPFS_ENABLED is set, so a node that does not host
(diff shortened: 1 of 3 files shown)


yashanand1910:
We will be using CSX to achieve this.

PR #872, opened 2026-09-15
CED-2346: Single source of truth for SLURM versions

gmarzjr:
## Describe your changes

The slurm version list was hardcoded in download.yml's fetch matrix and test_slurm.yml's test matrix, duplicating cedana-slurm/slurm-versions. The copies drifted whenever that file changed: removing a version made test_slurm.yml request an artifact nobody built ("Artifact not found: slurm-25-05-8-1-amd64"), and adding one meant it was built and published without ever being tested.

download.yml now reads the file in the cedana-slurm clone it was already doing, feeds its own matrix from it, and exposes the list as a workflow_call output. test_slurm.yml takes that list as an input and uses it directly as the `slurm` matrix dimension.

The first commit declares the test matrix where every version runs the full sweep. A follow-on commit ensures the test matrix is equivalent (in structure) to before this change, where one version of slurm is tested comprehensively with all tests and the others just run a CPU test. Once @TusharMohapatra07 ci optimizations are available to speed up CI runs, we could revert that follow-on commit if we wanted to run the tests across all slurm versions.

This is intended to resolve difficulty/blocking issues to support new SLURM versions such as in https://github.com/cedana/cedana-slurm/pull/170 .

This PR is required before the cedana-slurm PR can be approved/merged (https://github.com/cedana/cedana-slurm/pull/171)


.github/workflows/test_slurm.yml
```diff
@@ -45,16 +45,15 @@ on:
...
         type: string
-        description: "Name of the slurm binaries artifact to download"
-        required: false
-        default: "slurm"
+        description: "SLURM versions to test, from cedana-slurm's slurm-versions (JSON array)"
+        required: true
       slurm_default_version:
         type: string
         description: "SLURM version tag used by the single-version jobs"
         required: false
```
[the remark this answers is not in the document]
gmarzjr: the slurm version that is tested most comprehensively will be the first entry in the slurm-versions file in cedana-slurm now

.github/workflows/download.yml
```diff
@@ -426,21 +430,45 @@ jobs:
...
+
+      - name: Read slurm-versions
+        id: read
+        run: |
+          set -euo pipefail
+          # Entries are SchedMD git tags ("slurm-25-11-5-1") and are used verbatim.
+          versions=$(sed -E 's/#.*$//; s/^[[:space:]]*//; s/[[:space:]]*$//' slurm-versions | grep -v '^$')
+          json=$(jq -Rsc 'split("\n") | map(select(. != ""))' <<<"$versions")
+          echo "versions=$json" >> "$GITHUB_OUTPUT"
+          echo "Resolved SLURM versions: $json"
```
rahulk789, line 459: Could we possibly move this to a separate prepare.yml to stay consistent with with cedana-slurm if it makes sense.

gmarzjr: To me this slurm-versions job is downloading the slurm-versions file, which fits the pattern of this file to resolve external dependencies consumed by other workflows. And it is located right next to the slurm job that consumes its output which helps readability. But I'm open to starting a prepare.yml if you see that new file being semantically useful for other future jobs/workflow calls being placed there too. Feel free to push back.

yashanand1910: @rahulk789 `prepare.yml` is used in repos to prepare additional stuff (that belongs to that repo), that does not really belong in `build.yml`, just for the sake of organization (e.g. https://github.com/cedana/cedana-helm-charts/blob/66e84c27caa60ffd32c6f7cb5a8e0c3a22eec766/.github/workflows/prepare.yml). In this case, the `slurm-versions` don't really belong to this repo, and better placed where it's currently.

.github/workflows/test_slurm.yml
```diff
@@ -71,6 +65,41 @@ env:
   DEBUG: 0
 
 jobs:
+  # Restricts how much of the matrix below actually runs. Sweeping every SLURM version
```
yashanand1910, line 68: This is good.

yashanand1910 (approved):
Good work!

.github/workflows/download.yml
rahulk789: Understood. Thanks!

